PR #2, opened 2016-07-14
Setup travis ci config

pgoetze:
Set up CI with [Travis CI](https://travis-ci.org/iris-dni/iris-frontend)

* pushes to staging are deployed to the [iris-frontend-staging](http://iris-frontend-staging.herokuapp.com/) Heroku app
* pushes to master are deployed to the [iris-frontend](http://iris-frontend.herokuapp.com/) Heroku app


pgoetze:
ping @mattberridge

src/server.js
```diff
@@ -39,6 +43,19 @@ server.route({
 });
 
 /**
+ * Serve static bundled client.js, needed for production server
+ */
+server.route({
+  method: 'GET',
+  path: '/dist/client.js',
```
mattberridge: is there not a static assets setting in hapi, similar to `express.static` - or do you need to define a route for every file? 😕

package.json
```diff
@@ -20,10 +20,16 @@
     "ascii-json": "^0.2.0",
     "babel-core": "^6.10.4",
     "babel-loader": "^6.2.4",
+    "babel-plugin-transform-decorators-legacy": "^1.3.4",
+    "babel-plugin-typecheck": "^3.9.0",
```
mattberridge: this should be in `devDependencies`

webpack.config.prod.js
```diff
@@ -20,7 +20,8 @@ module.exports = {
     }),
     new webpack.DefinePlugin({
       'process.env': {
-        NODE_ENV: 'production'
+        // needs to be stringified, else it will be evaluated as a var production
+        NODE_ENV: JSON.stringify('production')
```
mattberridge: 👍

src/server.js
pgoetze: Oh yeah, sure, we can also configure a full directory.

package.json
pgoetze: Alright, will check whether deployment to Heroku works without them (Heroku is by default only installing the non-dev dependencies). I just put them there because they are also in the `.babelrc` config and deployment didn’t work without the preset modules at least.

mattberridge: only `babel-plugin-typecheck` I mean, not all

.babelrc
```diff
@@ -1,15 +1,17 @@
...
+    "production": {
+      "plugins": [
+        "transform-decorators-legacy"
+      ],
+      "presets": ["react", "es2015", "stage-0"],
+    },
+    "development": {
+      "plugins": [
+        "transform-decorators-legacy",
+        "typecheck"
```
mattberridge: ... because it's only used in the Dev environment

package.json
pgoetze: Ok, cool, makes sense.

PR #3, opened 2016-07-19
Test setup

pgoetze:
Sets  up a test environment with:

* [mocha](https://mochajs.org/) for unit tests
* [enzyme](http://airbnb.io/enzyme/) for component tests

The assertion lib is [chai](http://chaijs.com/).

We have basic example tests for routes, component, and reducers.

Test can be run with `npm run test` or `npm test`.
You can automatically watch code changes and run the tests with `npm run test:watch`

I assumed a name of `*.spec.js` for all the test files. I Found it to be used in the redux app examples.
This ensures that only actual test files are run (it excludes potential helper js files in the test directory) and makes it easier to jump to files when developing (compared to when all the files have the same name).

What do you think about this naming convention @mattberridge and @artibella?


test/routes.spec.js
```diff
@@ -0,0 +1,34 @@
+import chai from 'chai';
+import server from './../src/server';
+
+const assert = chai.assert;
```
mattberridge: const { assert } = chai

Tiny bit cleaner

test/reducers/counter.spec.js
```diff
@@ -0,0 +1,27 @@
+import chai from 'chai';
+import counter from '../../src/reducers/counter';
+
+const assert = chai.assert;
+
+describe('counter reducer', () => {
```
mattberridge: These are great! Really simple

test/components/Counter.spec.js
```diff
@@ -0,0 +1,50 @@
...
+    buttons.at(0).simulate('click');
+
+    assert.equal(actions.plus.calledOnce, true);
+  });
+
+  it('calls minus on clicking the second button', () => {
+    const { buttons, actions } = setup();
+    buttons.at(1).simulate('click');
+
+    assert.equal(actions.minus.calledOnce, true);
```
mattberridge: use the same format as below maybe?

assert.equal(actual, expected)

test/routes.spec.js
```diff
@@ -0,0 +1,34 @@
+import chai from 'chai';
+import server from './../src/server';
```
mattberridge: Does "clean" importing work in the test Env too? Webpack should look in the 'src' folder by default (it does in the app)

mattberridge: Ahh it's not complied by webpack, or?

pgoetze: It is. And it works indeed, thanks.

test/components/Counter.spec.js
pgoetze: Jep. Will change.

pgoetze: Hm, I think these can actually use `assert.isTrue` instead of `assert.equal`.

PR #6, opened 2016-07-21
[WIP] Webfonts and Theming

mattberridge:
This PR got out of hand again, need to break things up more (by task) in future. 
Here is a summary of what this PR contains

## Templating

Adds [Nunjucks](https://mozilla.github.io/nunjucks/) as a [templating engine](https://github.com/iris-dni/iris-frontend/blob/f2cd65fc40702818c7a876b1dd96c9be8ca6d355/src/server.js#L55), useful for our [index.html](https://github.com/iris-dni/iris-frontend/blob/f2cd65fc40702818c7a876b1dd96c9be8ca6d355/src/index.html) and includes for scripts and `<head>` from the project default and from the user's theme.

Added a [`copy:theme`](https://github.com/iris-dni/iris-frontend/blob/f2cd65fc40702818c7a876b1dd96c9be8ca6d355/scripts/copy-theme.js) node command to bundle the index.html file to the build folder.

## Env

Adds `.env` to the project with `dotenv` lib. Parsing [on the server](https://github.com/iris-dni/iris-frontend/blob/f2cd65fc40702818c7a876b1dd96c9be8ca6d355/babel.server.js#L2) added for server-side code. Also set up in the [webpack config](https://github.com/iris-dni/iris-frontend/blob/f2cd65fc40702818c7a876b1dd96c9be8ca6d355/webpack.config.prod.js#L1) for client bundle, with [`webpack.EnvironmentPlugin`](https://github.com/iris-dni/iris-frontend/blob/f2cd65fc40702818c7a876b1dd96c9be8ca6d355/webpack.config.prod.js#L39) to define vars in `process.env` globally. We keep the `webpack.DefinePlugin` in production below this, to set the production mode for React via `process.env.NODE_ENV`.

The only `.env` var so far is `THEME_PATH` where the user defines their theme folder.

Added [`.env.default`](https://github.com/iris-dni/iris-frontend/blob/f2cd65fc40702818c7a876b1dd96c9be8ca6d355/.env.default) for the default settings checked-in to the repo.

## Config

I created a [master config file](https://github.com/iris-dni/iris-frontend/blob/f2cd65fc40702818c7a876b1dd96c9be8ca6d355/src/config/index.js) which will export the config for the app.

It uses a [`mergeConfig`](https://github.com/iris-dni/iris-frontend/blob/f2cd65fc40702818c7a876b1dd96c9be8ca6d355/src/config/merge.js) helper to `Object.assign` the default and the custom (this will need extending in future I am sure, with `deepMerge` or so).

I also created a [`requireThemeFile`](https://github.com/iris-dni/iris-frontend/blob/f2cd65fc40702818c7a876b1dd96c9be8ca6d355/src/helpers/requireThemeFile.js) helper where it grabs the theme path from `.env` and tries to require a file (JS or JSON) from the theme. If the file does not exist or errors then it returns an empty object. This is meant for requiring files which will overwrite defaults in the app if the user has provided them.

So now we have some theming flexibility with a [default theme config](https://github.com/iris-dni/iris-frontend/blob/f2cd65fc40702818c7a876b1dd96c9be8ca6d355/src/theme/config.js) in the repo and then optional overwritten config defined by the user in the theme, merged and required in the app via `import config from 'config';`.

Config also includes an abstracted [`getBundles`](https://github.com/iris-dni/iris-frontend/blob/f2cd65fc40702818c7a876b1dd96c9be8ca6d355/src/config/bundles.js) function which define the core (i.e. required) script and stylesheet includes for the app, looped [in the template](https://github.com/iris-dni/iris-frontend/blob/f2cd65fc40702818c7a876b1dd96c9be8ca6d355/src/index.html#L10). 

## Theme

- The default theme for the app is now developed in `/src/theme`
- The user defines their own, separate theme path in `.env` and customises their theme there

I then added a [`copy-theme`](https://github.com/iris-dni/iris-frontend/blob/f2cd65fc40702818c7a876b1dd96c9be8ca6d355/scripts/copy-theme.js) node script to merge the two together and output to the build folder. Any files the user overwrites in their custom theme are taken and all additional ones added (for assets etc.)

An example of this is [`/theme/includes/scripts.html`](https://github.com/iris-dni/iris-frontend/blob/f2cd65fc40702818c7a876b1dd96c9be8ca6d355/src/theme/includes/scripts.html) as a partial where the user will add custom tracking etc. scripts. By default we provide a blank file referenced [in the template](https://github.com/iris-dni/iris-frontend/blob/f2cd65fc40702818c7a876b1dd96c9be8ca6d355/src/index.html#L24) - which would be overwritten by a file of the same name in the user's theme

## Webfonts

We include a [webfonts partial](https://github.com/iris-dni/iris-frontend/blob/f2cd65fc40702818c7a876b1dd96c9be8ca6d355/src/index.html#L9) in the default theme which includes the default font of Open Sans via Google and loading [via `loadCss` method](https://github.com/iris-dni/iris-frontend/blob/f2cd65fc40702818c7a876b1dd96c9be8ca6d355/src/theme/includes/webfonts.html). We keep flexibility here as the user (and us, later) can customise this include in their theme folder with the preferred / suitable [font loading method](https://www.zachleat.com/web/comprehensive-webfonts/).

Font files / assets in the user's theme folder will be copied over via the build step above.

## Assets

Static assets and scripts / styles which are core to the app (and not intended for customising) now live in the `/src/assets` folder.


src/server/render.js
```diff
@@ -27,10 +23,14 @@ export default (request, reply) => {
...
-      const webserver = process.env.NODE_ENV === 'production' ? '' : '//localhost:8080';
+      const basePath = process.env.NODE_ENV === 'production' ? '' : '//localhost:8080';
 
-      reply(markup(reactString, stateJSON, webserver, config));
+      reply.view('index', {
+        reactMarkup: reactString,
+        initialState: JSON.stringify(initialState),
+        javascripts: [`${basePath}/dist/client.js`],
+        stylesheets: process.env.NODE_ENV === 'production' ? ['/dist/styles.css'] : ''
+      });
```
mattberridge: @pgoetze let's think about how we handle all the 'markup' varibles, especially when combined with over-rides / additions from the user's theme

package.json
```diff
@@ -6,7 +6,7 @@
   "repository": "https://github.com/iris-dni/iris-frontend",
   "scripts": {
     "clean": "rm -rf static/dist",
-    "nodecheckversion": "node ./scripts/nodecheckversion.js",
+    "nodecheckversion": "node ./scripts/node-checkversion.js",
```
HugoGiraudel: For what it is worth, I am against a npm scripts setup. I’d favor Gulp.

scripts/copy-theme.js
```diff
@@ -0,0 +1,23 @@
+/* eslint-disable */
+var fs = require('fs-extra');
+require('dotenv').config();
+
+var DEFAULT_THEME_PATH = 'src/theme';
```
HugoGiraudel: `./src/theme` seems more explicit.

scripts/copy-theme.js
```diff
@@ -0,0 +1,23 @@
...
+var CUSTOM_THEME_PATH = process.env.THEME_PATH;
+var TARGET_PATH = 'static/dist/theme';
```
HugoGiraudel: `./static/dist/theme` seems more explicit.

src/config/bundles.js
```diff
@@ -0,0 +1,19 @@
+import getIncludedAssetPath from 'helpers/getIncludedAssetPath';
+
+export default () => {
+  let stylesheets;
```
HugoGiraudel: `let stylesheets = [];`

src/config/bundles.js
```diff
@@ -0,0 +1,19 @@
...
+
+  const javascripts = [
+    getIncludedAssetPath('client.js')
+  ];
+
+  if (process.env.NODE_ENV === 'production') {
+    stylesheets = [
```
HugoGiraudel: `stylesheets.push(getIncludedAssetPath('styles.css'))`

src/config/bundles.js
```diff
@@ -0,0 +1,19 @@
...
+      getIncludedAssetPath('styles.css')
+    ];
+  } else {
```
HugoGiraudel: Remove.

src/config/index.js
```diff
@@ -0,0 +1,13 @@
...
+import getBundles from 'config/bundles';
+import requireThemeFile from 'helpers/requireThemeFile';
+
+const customConfig = requireThemeFile('config');
+const mergedConfig = mergeConfig(defaultConfig, customConfig);
+const bundles = getBundles();
+
+export default Object.assign({},
+  mergedConfig,
+  { bundles: bundles }
```
HugoGiraudel: `{ bundles }`

src/helpers/getIncludedAssetPath.js
```diff
@@ -0,0 +1,7 @@
+export default (filePath) => {
+  const basePath = process.env.NODE_ENV === 'production'
+    ? ''
+    : '//localhost:8080';
+
+  return `${basePath}/dist/${filePath}`;
```
HugoGiraudel: `return [basePath, 'dist', filePath].join('/')`

src/helpers/requireThemeFile.js
```diff
@@ -0,0 +1,19 @@
+export default (filePath) => {
+  const themePath = process.env.THEME_PATH;
+
+  if (!themePath) {
+    return {};
+  }
+
+  const requirePath = `../../${themePath}/${filePath}`;
```
HugoGiraudel: `path.join('..', '..', themePath, filePath)`

src/helpers/requireThemeFile.js
```diff
@@ -0,0 +1,19 @@
...
+
+  try {
+    const loadedFile = require(requirePath);
+    if (typeof loadedFile === 'object' && loadedFile.default) {
```
HugoGiraudel: Why do you do this?

HugoGiraudel: `return loadedFile.default || {}` should be enough.

src/index.html
```diff
@@ -0,0 +1,22 @@
+<!doctype html>
+<html lang="{{ lang }}">
```
HugoGiraudel: You could `| default: "en"`, just in case.

src/index.html
```diff
@@ -0,0 +1,22 @@
...
+  <head>
+    <meta charset="{{ charset }}">
```
HugoGiraudel: You could `| default: "utf-8"`, just in case.

src/theme/includes/webfonts.html
```diff
@@ -0,0 +1,6 @@
+<script>
+  {% include 'assets/scripts/loadCss.js' %}
+</script>
+<script>
+  loadCSS('https://fonts.googleapis.com/css?family=Open+Sans:400,700');
```
HugoGiraudel: Best would be to have to WOFF and WOFF2 files locally?

test/config/bundles.js
```diff
@@ -0,0 +1,38 @@
+import chai from 'chai';
```
HugoGiraudel: Is there any reason you use `chai` rather than Node.js `assert`?

mattberridge: As @pgoetze set it up that way in another PR, because of a standardising the test setup across projects (as discussed with @artibella)

HugoGiraudel: I personally don’t think chai is necessary for testing, and Node.js assert library is more than enough.

src/theme/includes/webfonts.html
mattberridge: Keeping things simple for now in the default theme. Google recommends using the embed code /and serves `woff` / `woff2` depending on the browser used.

PR #7, opened 2016-07-26
[WIP] Provide initial state

mattberridge:


src/containers/Petition.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { connect } from 'react-redux';
+
+const Petition = (props) => (
```
sigmarstern: According to [this](https://github.com/airbnb/javascript/tree/master/react#class-vs-reactcreateclass-vs-stateless), shouldn't it read

``` 
function Petition({ props }) {
  return <div>…</div>;
}
```

PR #12, opened 2016-07-29
[RFR] Petition Detail Page

sigmarstern:
This PR is about adding the `/petitions` routing and async data fetching for petition data by id. Large restructuring of the app took place to get Redux working with async data, in line with [this excellent article](https://blog.tableflip.io/server-side-rendering-with-react-and-redux/).

In short:
- On the server: we only render the page once the handler component's `fetchData()` function's Promises have resolved.
- The data hydrates the Redux store and is passed to the initial state
- In the client: we call the same fetch action on `componentDidMount()` to fetch the data if not provided / hydrated by initial state, i.e. in the case of navigation after first-load.

I have added comments below to inform in a bit more detail.


src/server/getBrowserTitle.js
```diff
@@ -0,0 +1,19 @@
...
+  // Petitions route
+  if (pathname && pathname.indexOf('/petitions') > -1) {
+    const { petition } = state;
+    titlePrefix = petition && petition.title;
+  }
+
+  return titlePrefix
+    ? TITLE_TEMPLATE.replace('%s', titlePrefix)
+    : config.title;
+};
```
mattberridge: I am still a bit unsure of this file and how we proceed with more routes. It's concerned with getting the page title for each 'page type' and returning for the server-side markup (added in Nunjucks).

This happens before we've started the React render tree so we get the location of the request and the full state (passed from the store), then match the route with the title we want - here `petition.title`.

src/helpers/requireThemeFile.js
```diff
@@ -7,10 +7,10 @@ export default (filePath) => {
     return {};
   }
 
-  const requirePath = path.join('..', '..', themePath, filePath);
+  const requirePath = path.join('/', themePath, filePath);
 
   try {
-    const loadedFile = require(requirePath);
+    const loadedFile = require('../..' + requirePath);
```
mattberridge: Webpack gives a warning if you just use a variable for requires, and no parts of a string  ¯\_(ツ)_/¯

src/routes.js
```diff
@@ -1,14 +1,29 @@
...
+export default function (props = {}) {
+  let history = browserHistory;
+
+  if (props.store) {
+    history = syncHistoryWithStore(browserHistory, props.store);
+  }
+
+  return (
+    <Router history={history}>
+      <Route path='/' component={Layout}>
```
mattberridge: This route wraps our app, and provides us with a way to add header nav, footer etc.

src/views/Layout.js
```diff
@@ -0,0 +1,11 @@
+import 'assets/styles/reset.css';
+import React from 'react';
+import Helmet from 'react-helmet';
+import { TITLE_TEMPLATE } from 'server/getBrowserTitle';
+
+export default ({ children }) => (
+  <div>
+    <Helmet titleTemplate={TITLE_TEMPLATE} />
```
mattberridge: Here we globally define the default template for that page `<title>`, via the template exported in the `getBrowserTitle` helper.

src/containers/Petition.js
```diff
@@ -0,0 +1,54 @@
...
+  componentDidMount () {
+    if (!this.props.id) {
+      this.props.fetchPetition(this.props.params.id);
+    }
+  },
+
+  render () {
+    return (
+      <div>
+        <Helmet title={this.props.title} />
```
mattberridge: Defines the page `<title>` added to the title template for this page type, i.e. the petition title

src/containers/Petition.js
```diff
@@ -0,0 +1,54 @@
...
+import { connect } from 'react-redux';
+import { fetchPetition } from 'actions/PetitionActions';
+import Petition from 'components/Petition';
+
+const PetitionContainer = React.createClass({
+  // When the component gets added to the DOM,
+  // fetch Petition if `id` is not defined.
+  componentDidMount () {
+    if (!this.props.id) {
+      this.props.fetchPetition(this.props.params.id);
```
mattberridge: This gets the petition by URL param when we navigate client-side. It is not run if a petition `id` is given i.e. by the initial state on first load.

src/containers/Petition.js
```diff
@@ -0,0 +1,54 @@
...
+        <Helmet title={this.props.title} />
+        <Petition {...this.props} />
+      </div>
+    );
+  }
+});
+
+PetitionContainer.fetchData = ({ store, params }) => {
+  return store.dispatch(fetchPetition(params.id));
+};
```
mattberridge: This is the key for of server-side redux rendering. In `server/render.js` we grab this function and dispatch the `fetchPetition` action, to hydrate the store when the route hits. The page isn't rendered until the promise is fulfilled.

src/server/render.js
```diff
@@ -1,31 +1,65 @@
...
+      const components = renderProps.components;
+      // Extract our page component
+      const Comp = components[components.length - 1].WrappedComponent;
+      // Extract `fetchData` if exists
+      const fetchData = (Comp && Comp.fetchData) || (() => Promise.resolve());
+      // Get from renderProps
+      const { location, params, history } = renderProps;
+
+      // Fetch async data, then render
+      fetchData({ store, location, params, history })
```
mattberridge: Here we are using our `fetchData` function obtained from the component lookup above (i.e. the handler given from the routing). We carry out any async requests and `then()` to render the page when all promises are fulfilled.

PR #19, opened 2016-08-04
Petitions

sigmarstern:
Added /petitions route with basic limit/offset and per/page pagination.

@mattberridge, maybe you can have a look into it and correct me where I'm wrong. For instance pagination links on `/petitions` list aren't working properly while they are on `/`. I guess there is a more appropriate way of passing those pagination params back and forth.


src/containers/Petitions.js
```diff
@@ -0,0 +1,50 @@
...
+  total: React.PropTypes.number,
+  page: React.PropTypes.number,
+  per: React.PropTypes.number,
+  fetchPetitions: React.PropTypes.func
+};
+
+const mapStateToProps = ({ petitions }) => ({
+  petitions: petitions.data || [],
+  total: petitions.total,
+  page: petitions.page,
```
mattberridge: think this could be better as `currentPage` for clarity

src/containers/Petitions.js
```diff
@@ -0,0 +1,50 @@
...
+  per: petitions.per
```
mattberridge: and this as `perPage`

src/components/Petitions/index.js
```diff
@@ -0,0 +1,19 @@
...
+import Pager from 'components/Pager';
+
+const Petitions = ({ total, petitions, page, per }) => (
+  <div>
+    <ul>
+      {petitions.map((petition, index) => {
+        return (
+          <PetitionListItem {...petition} key={petition.id} />
+        );
+      })}
```
mattberridge: here you can simplify to

`{petitions.map((petition, index) => <PetitionListItem {...petition} key={petition.id} />)}`

src/components/Petitions/index.js
```diff
@@ -0,0 +1,19 @@
...
+    </ul>
+
+    <Pager {...{ path: '/petitions', page: page, per: per, total: total }} key={page} />
```
mattberridge: prefer

```
<Pager
  path: '/petitions'
  page: page
  per: per
  total: total
/>
```

don't think we need `key`? that's just if you are iterating over components :)

src/components/Petitions/index.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import PetitionListItem from 'components/PetitionListItem';
+import Pager from 'components/Pager';
+
+const Petitions = ({ total, petitions, page, per }) => (
```
mattberridge: `{ total, petitions, currentPage, perPage }`

PR #21, opened 2016-08-10
Feature/145 finalize petitions

sigmarstern:
@mattberridge:

- Added missing page title "Petitionen"
- Added headline "Petitionen"

### DIFF
--- a/src/components/Petitions/index.js
+++ b/src/components/Petitions/index.js
@@ -1,9 +1,10 @@
 import React from 'react';
 import PetitionListItem from 'components/PetitionListItem';
+import settings from 'settings';
 
 const Petitions = ({ total, petitions, currentPage, perPage }) => (
   <div>
-    <h1>Petitionen</h1>
+    <h1>{settings.petitionsText}</h1>
     <ul>
       {petitions.map((petition) => <PetitionListItem {...petition} key={petition.id} />)}
     </ul>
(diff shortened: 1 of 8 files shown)


src/components/Petitions/index.js
```diff
@@ -2,9 +2,12 @@ import React from 'react';
 import PetitionListItem from 'components/PetitionListItem';
 
 const Petitions = ({ total, petitions, currentPage, perPage }) => (
-  <ul>
-    {petitions.map((petition) => <PetitionListItem {...petition} key={petition.id} />)}
-  </ul>
+  <div>
+    <h1>Petitionen</h1>
```
mattberridge: Ideally should come from [`config.js`](https://github.com/iris-dni/iris-frontend/blob/develop/src/settings/config.js) for now, before we add translation handling. Give as 'Petitions' in English for IRIS, then we over-write for Petitio.ch in the custom config.

src/containers/Petitions.js
```diff
@@ -12,7 +12,7 @@ const PetitionsContainer = React.createClass({
   render () {
     return (
       <div>
-        <Helmet title='Petitions' />
+        <Helmet title='Petitionen' />
```
mattberridge: Same here. Plus this only does the client-side page title. The server-side page is defined in https://github.com/iris-dni/iris-frontend/blob/develop/src/server/getBrowserTitle.js :)
e.g.
```
    case 'PetitionsContainer':
      return TITLE_TEMPLATE.replace('%s', settings.petitionsTitle);
```

sigmarstern: Thanks! Will do.

PR #25, opened 2016-08-12
[RFR] Petitioner Login

sigmarstern:
- Add SSO workflow
- Add an example route `/restricted` which can be removed later
- Refactored API client and repositories

Ping @mattberridge, @artibella


src/services/api/client.js
```diff
@@ -30,7 +26,8 @@ export default {
     return axios({
       method: method,
       url: apiUrl(requestPath),
-      data: data
-    });
+      data: data,
+      withCredentials: true
+    }).then(response => (response.data));
```
mattberridge: tiny but I think our preferred convention would be

`.then((response) => response.data);`

sigmarstern: Sure. Fixed this.

src/services/auth/Login.js
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+import { withRouter } from 'react-router';
+import ssoProviders from 'settings/ssoProviders';
```
mattberridge: this needs to be imported directly from `settings` so the user can over-write it in their theme

PR #26, opened 2016-08-12
Add new circular progress bar to petition teasers

artibella:
This one got a liltte bigger - sorry!

* Adds a nice litte circle with the progress to the petition teasers - uses [ProgressBar.js](https://kimmobrunfeldt.github.io/progressbar.js)
Preview
* refactored the `Dial` component to use it and renamed it to `PetitionMetric` to make it more generic.
* hide the city in the petition teasers if not set
* added colors to global config object. Currently this is duplicating the color values from the `_colors.scss` file which is not so nice, but works for now


src/components/PetitionMetric/index.js
```diff
@@ -0,0 +1,31 @@
...
+          <dd className={styles.figure}>
+            {figure}
+          </dd>
+        </dl>
+      </div>
+    );
+  }
+});
+
+export default PetitionMetric;
```
mattberridge: you could make this a pure, stateless component :)

```
import React from 'react';
import styles from './petitionmetric.scss';
import CircleProgressBar from 'components/CircleProgressBar';
import Icon from 'components/Icon';

export default ({ icon, figure, caption, percentage }) => {
  return (
    <div className={styles.root}>
      <div className={styles.progress}>
        <CircleProgressBar percentage={percentage} size='small'>
          <Icon id={icon} size='small' />
        </CircleProgressBar>
      </div>
      <dl className={styles.text}>
        <dt className={styles.caption}>
          {caption}
        </dt>
        <dd className={styles.figure}>
          {figure}
        </dd>
      </dl>
    </div>
  );
};
```

src/selectors/petitionTeaser.js
```diff
@@ -16,14 +16,14 @@ export default (petition) => {
...
       info: {
-        city: petition.city || 'Aargau',
+        city: petition.city,
         owner: getPetitionAuthor(petition.owner || {})
       },
-      dialTime: {
+      timeMetric: {
         figure: daysRemaining,
-        percentage: calculatePercentage(daysRemaining, settings.daysToVote)
+        percentage: 100 - calculatePercentage(daysRemaining, settings.daysToVote)
```
mattberridge: not sure if this should / could be baked-in to the helper somewhere? thoughts?

src/components/CircleProgressBar/index.js
```diff
@@ -0,0 +1,65 @@
...
+    };
+  },
+
+  componentDidMount () {
+    this.initProgressBar();
+    this.updateProgressBar(this.props.percentage);
+  },
+
+  initProgressBar () {
+    let canvasElement = ReactDOM.findDOMNode(this.refs.canvas);
```
mattberridge: `const`

src/components/CircleProgressBar/index.js
```diff
@@ -0,0 +1,65 @@
...
+      svgStyle: {
+        width: '100%',
+        display: 'block',
+        strokeLinecap: 'round'
+      }
+    });
+  },
+
+  updateProgressBar (percentage) {
+    let progress = percentage / 100;
```
mattberridge: `const`

src/components/CircleProgressBar/index.js
```diff
@@ -0,0 +1,65 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import styles from './circleprogressbar.scss';
+import settings from 'settings';
+import ProgressBar from 'progressbar.js';
+
+const CircleProgressBar = React.createClass({
+
+  getDefaultProps () {
```
mattberridge: cleaner version

```
getDefaultProps: () => ({
 ...
}),
```

:)

src/components/CircleProgressBar/index.js
```diff
@@ -0,0 +1,65 @@
...
+  },
+
+  render () {
+    return (
+      <div className={styles.root}>
+        <div className={styles[this.props.size]} ref='canvas'></div>
+        <div className={styles.content}>
+          {this.props.children}
+        </div>
+      </div>
```
mattberridge: Not sure on the semantics of these elements. Everything a `<div>` doesn't seem right

src/components/TeaserFooter/index.js
```diff
@@ -1,26 +1,26 @@
...
-        <Dial
-          {...dialSupporters}
-          icon={'signature'}
+      <div className={styles.metric}>
+        <PetitionMetric
+          {...supportersMetric}
+          icon={'Signature'}
           caption={settings.teaserSupportersText}
         />
       </div>
```
mattberridge: Despite what I said before, I might break this down further after seeing it now.

```
<div className={styles.metrics}>
  <TeaserMetrics {...metrics} />
</div>
```

mattberridge:
Really good stuff! Good to see the finer parts of the Teasers coming together.

For reference, here's the library used which works excellently:

https://kimmobrunfeldt.github.io/progressbar.js/

src/selectors/petitionTeaser.js
artibella: I think it is ok for now as it is. If we need that logic more often we can bake it in

artibella:
Also now enhanced the Countdown component to use the new circular progress bar

PR #27, opened 2016-08-16
Add tests for the petitionEndDate selector

artibella:


test/selectors/petitionEndDate.js
```diff
@@ -0,0 +1,32 @@
+import chai from 'chai';
+import moment from 'moment';
+import getPetitionEndDate from 'selectors/petitionEndDate';
+import settings from 'settings';
+
+const { assert } = chai;
```
sigmarstern: Can't we use `import { assert } from 'chai';` instead?

PR #28, opened 2016-08-16
Feature/petitions api

sigmarstern:
@mattberridge, @artibella: Added missing functions to petition repository:

- `create(petition)`
- `update(petition)`
- `publish(petition)`

Also added some more tests.


src/services/api/client.js
```diff
@@ -14,19 +14,26 @@ const apiUrl = (requestPath) => {
 
 export default {
   request: (requestPath = '', data = {}, method = 'GET') => {
-    if (method === 'GET') {
-      const requestParams = queryParams.encode(data);
+    let payload = {};
+    let requestParams = queryParams.encode(data);
```
artibella: Maybe use `const` if the value does not change. But I'm not too strict on this 😏

artibella:
👍  Nice work and good test coverage!

PR #30, opened 2016-08-17
Hide progress if petition has no required supporters 

artibella:
It is possible to submit petitions without a city. Therefore it won't be possible to display the supporters progress as the required number is defined on city level.

This PR does following:

* Hide supporters progress in teasers
* Hide supporters progress on petition page
* Refactor progress bar to use same logic as circular progress bar
* Refactor petition selector to use petition metrics object for more consistency

### DIFF
--- a/src/components/PetitionHeader/index.js
+++ b/src/components/PetitionHeader/index.js
@@ -5,18 +5,23 @@ import PetitionInfo from 'components/PetitionInfo';
 import PetitionStats from 'components/PetitionStats';
 import ProgressBar from 'components/ProgressBar';
 
-const PetitionHeader = ({ title, info, supporters, percentComplete }) => (
+const PetitionHeader = ({ title, info, metrics }) => (
   <header className={styles.root}>
     <div className={styles.heading}>
       <Heading1 text={title} />
     </div>
     <div className={styles.info}>
       <PetitionInfo {...info} />
-      <ProgressBar
-        animate
-        percentage={percentComplete}
-      />
-      <PetitionStats {...supporters} />
+      {metrics.supportersMetric.votingActive &&
+        <div className={styles.progress}>
+          <ProgressBar
+            animated
+            percentage={metrics.supportersMetric.percentage}
+            aria={metrics.supportersMetric.aria}
+          />
+          <PetitionStats {...metrics.supportersMetric} />
+        </div>
+      }
     </div>
   </header>
 );
--- a/src/components/PetitionHeader/index.js
+++ b/src/components/PetitionHeader/index.js
@@ -12,15 +12,17 @@ const PetitionHeader = ({ title, info, metrics }) => (
     </div>
     <div className={styles.info}>
       <PetitionInfo {...info} />
-      {metrics.supportersMetric.votingActive &&
-        <div className={styles.progress}>
-          <ProgressBar
-            animated
-            percentage={metrics.supportersMetric.percentage}
-            aria={metrics.supportersMetric.aria}
-          />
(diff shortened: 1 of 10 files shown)


src/selectors/petitionMetrics.js
```diff
@@ -13,6 +13,7 @@ export default (petition) => {
   const amountVotes = petition.supporters.amount;
   const timePercentage = getTimePercentage(daysRemaining, daysToVote);
   const votesPercentage = calculatePercentage(amountVotes, requiredVotes);
+  const votingActive = requiredVotes > 0;
```
artibella: I left this intentionally to `> 0` as the API does not provide a `-1` value yet. Should be changed later

src/selectors/petitionMetrics.js
```diff
@@ -33,7 +34,8 @@ export default (petition) => {
         minimum: 0,
         maximum: requiredVotes,
         value: amountVotes
-      }
+      },
+      votingActive: votingActive
```
sigmarstern: Consider having a test for this in `test/selectors/petitionMetrics.js`.

src/components/PetitionHeader/index.js
```diff
@@ -5,18 +5,23 @@ import PetitionInfo from 'components/PetitionInfo';
...
       <Heading1 text={title} />
     </div>
     <div className={styles.info}>
       <PetitionInfo {...info} />
-      <ProgressBar
-        animate
-        percentage={percentComplete}
-      />
-      <PetitionStats {...supporters} />
+      {metrics.supportersMetric.votingActive &&
```
mattberridge: `metrics.supportersMetric && metrics.supportersMetric.votingActive &&` for safety I think :)

src/selectors/petitionMetrics.js
mattberridge: I was seeing this lying outside the `metric` key, at the base of the object. but it doesn't matter I guess, we can always move it later if another instance (not related to metrics) is needed

mattberridge:
Incidentally, I found this before:
https://blog.risingstack.com/react-js-best-practices-for-2016/#keepyourstateflat

Saying it's recommended to keep your state state and use `normalizr`

src/selectors/petitionMetrics.js
artibella: Good idea

artibella: Added tow tests with commit a0e89d2

test/selectors/petitionMetrics.js
```diff
@@ -47,5 +47,21 @@ describe('get petiton metrics', () => {
...
     });
+
+    it('voting is enabled if required supporters above 0', () => {
+      const actual = metrics.supportersMetric.votingActive;
+      assert.isTrue(actual);
+    });
+
+    it('voting is disabled if required supporters below 0', () => {
+      const testPetiton = Object.assign({}, petition);
+      let metrics;
```
mattberridge: any reason to define then reassign?

test/selectors/petitionMetrics.js
```diff
@@ -47,5 +47,21 @@ describe('get petiton metrics', () => {
...
+
+      testPetiton.supporters.required = -1;
+      metrics = getPetitionMetrics(testPetiton);
```
mattberridge: `const metrics = getPetitionMetrics(testPetiton);`

mattberridge:
let's merge :) super

test/selectors/petitionMetrics.js
artibella: just old habits :)

define variable blocks first and then do the magic

PR #31, opened 2016-08-18
[RFR] Redux Containers as Child Components

mattberridge:
After looking at the bug on `develop` : `TypeError: Cannot read property 'timeMetric' of undefined` I realised we have already some problems with the data structure and passing down props.

**Basically, we missed half of the core concept of Redux!**

## Container Components

[Read more here](https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0#.58sncpld1). Until fairly recently it was advised in the React/Redux docs to only use one container component high up in the tree (per route essentially), and then pass props down.

This is actually not the case, it's *far better* to subscribe containers to the Redux Store as and when you need them in (child) components instead of passing lots of props down the tree.

In this PR you can see I [replaced `PetitionSidebar`](https://github.com/iris-dni/iris-frontend/pull/31/files#diff-430ed53a934411fe66a682b9633808e0R8) with a new container component which [doesn't require any props](https://github.com/iris-dni/iris-frontend/pull/31/files#diff-430ed53a934411fe66a682b9633808e0R19) from above it in the tree.

Inside this container we [subscribe to the Redux store](https://github.com/iris-dni/iris-frontend/blob/4c3fa9f697d6293f905977064c9860474c720656/src/containers/PetitionSidebar.js#L20) (which already contains the `petition` state because of the `PetitionContainer` further up the tree), get the `petition` and [`mapStateToProps`](https://github.com/iris-dni/iris-frontend/blob/4c3fa9f697d6293f905977064c9860474c720656/src/containers/PetitionSidebar.js#L10).

It is here we use our [`petitionTimeMetric` selector](https://github.com/iris-dni/iris-frontend/blob/4c3fa9f697d6293f905977064c9860474c720656/src/selectors/petitionTimeMetric.js) to manipulate the full petition into the shape we need for this component.

That's our data mapping done (always in a _Container_) so we go ahead and render the **Presentational** [`PetitionSidebar` component](https://github.com/iris-dni/iris-frontend/pull/31/files#diff-b610b03e13aa726c9443e845d9986f9cR7) which simply takes and uses our mapped props.

## Selectors

The `/selectors` folder contains (pure) functions which take part of the state (e.e. the `petition`) and return an even smaller part of it. Before this was just being used at the root container e.g. `PetitionContainer` within the overall Petition selector, basically as a helper function.

A selector's intention is actually to take part of the state and just *select* the bits you need for / in _any_ component. So we should create selectors for any part of our app which uses data and select it that way. Rather than selecting it all up front and passing it down to child components. Winner.

An example of a small selector is [`petitionTimeMetric`](https://github.com/iris-dni/iris-frontend/blob/4c3fa9f697d6293f905977064c9860474c720656/src/selectors/petitionTimeMetric.js).

Selectors should only be used in Container Components, when mapping state to props (in the `mapStateToProps` function).

Many selectors are also easily testable with only one mock data file.


eschaefer:
Hey this is excellent Matt. Will probably reference this in the future.

sigmarstern:
@mattberridge, @artibella Rebased this with develop. Should be able to merge now.

PR #32, opened 2016-08-18
[RFR] Refactor Authentication Logic

sigmarstern:
@mattberridge, @artibella:

Add actions/reducers to the `LoginLink` component in order to dispatch the `whoAmI` API requests properly. Fixes the issue that caused this warning:

```
Warning: setState(...): Can only update a mounting component. This usually means you called setState() outside componentWillMount() on the server. This is a no-op. Please check the code for the LoginLink component.
```

Edit:

I Did some more refactorings and ended up having a top-level `App` container that currently takes care of prepopulating `props.me` but might also handle I18n and other concerns later.

There is now a higher-order component named `Restricted`, which checks for `props.me` and either renders restricted content or redirects to login page.

`Login` and `Logout` pages are regular containers now.

Also restructured settings for `ssoProviders` and the default `afterLoginPath`/`afterLogoutPath`.


mattberridge:
Awesome work! Was hoping we could do this 👍 
I think you might be able to use this to change `LoginLink` to be a much simpler component - just subscribing to `me` from the global state instead of doing its own `whoAmI` request?

src/containers/App.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import { fetchWhoAmI } from 'actions/AuthActions';
+import Layout from 'views/Layout';
+
+const App = React.createClass({
+  getInitialState: () => ({ me: false }),
```
mattberridge: do we need this if we aren't using state in the component?

src/settings/index.js
```diff
@@ -11,7 +11,11 @@ const customConfig = needsCustomThemePath
...
 
-export default Object.assign({},
+const all = Object.assign({},
   mergedSettings,
   { bundles }
 );
+
+export default all;
+export const ssoProviders = all.ssoProviders;
+export const authSettings = all.auth;
```
mattberridge: cool idea to export!

mattberridge:
Great stuff. Loving how small and simple each component is!

PR #39, opened 2016-08-25
[RFR] Sprint 3 cleaning

mattberridge:
Changes / amends following sprint 3 review


src/components/CircleProgressBar/index.js
```diff
@@ -17,19 +17,32 @@ const CircleProgressBar = React.createClass({
...
+
+  componentDidUpdate (prevProps) {
+    if (prevProps.percentage !== this.props.percentage) {
+      this.updateProgressBar(this.props.percentage);
+    }
+  },
+
+  componentWillUnmount () {
+    this.progressBar.destroy();
   },
```
mattberridge: ping @artibella - needed to destroy the JS on unmount

src/components/CircleProgressBar/index.js
```diff
@@ -17,19 +17,32 @@ const CircleProgressBar = React.createClass({
 
   componentDidMount () {
     this.initProgressBar();
-    this.updateProgressBar(this.props.percentage);
+    if (this.props.percentage > 0) {
+      this.updateProgressBar(this.props.percentage);
+    }
```
mattberridge: ping @artibella - only init the bar if `percentage > 0` (stop weird flash of the line then disappearing when equals 0)

src/components/CircleProgressBar/index.js
```diff
@@ -17,19 +17,32 @@ const CircleProgressBar = React.createClass({
...
+  },
+
+  componentDidUpdate (prevProps) {
+    if (prevProps.percentage !== this.props.percentage) {
+      this.updateProgressBar(this.props.percentage);
+    }
```
mattberridge: ping @artibella - update the progress bar when the component updates (i.e when we fetch a new petition from the state)

src/components/ProgressBar/index.js
```diff
@@ -16,7 +16,19 @@ const ProgressBar = React.createClass({
...
+  },
+
+  componentDidUpdate (prevProps) {
+    if (prevProps.percentage !== this.props.percentage) {
+      this.updateProgressBar(this.props.percentage);
+    }
+  },
+
+  componentWillUnmount () {
+    this.progressBar.destroy();
```
mattberridge: ping @artibella - all this logic is duplicated. would be nice to abstract a generic `ProgressElement` with options for circle, line, size etc. so we don't have this  in two places.

PR #41, opened 2016-08-25
[WIP] Sign-in Overlay

mattberridge:
## Outcome

Create a modal-type box which contains the sign-in actions. We will then re-use the same functionality for when a user tries to support a Petition but are not authenticated

screen shot 2016-09-01 at 12 33 57

## Technical

- Added `ModalWindow` redux actions and reducer with container and component
- Added `<ModalTrigger>` container + component with `authenticating` flag and `modal` and `href` props. Triggers `onClick` to show `modal` if user is not signed in, follows `href` if they are.
- Switch-like statement inside `<ModalWindowContainer>` selects component to display given `modalWindow{ type: '...' }` state
- Helper to allow `returnUrl` to be obtained in reducer from `react-router` `location`, for use in Sso links
- Refactors Sso logic and components, with helpers
- Adds `modal-active` class to `<html>` when modal active
- Adds `classList` polyfill for the above
- Further customisation options for `<Button>`
- Further customisation options for `<ButtonLink>` including `external` prop to disable React routing
- Adds `<ButtonIcon>` component for adding icons inside buttons
- Adds `<CheckAuth>` wrapper to show loading screen (design TBD) when we have the `isLoading` flag set in the `me` object
- Adds generic `<CloseButton>` component, with `onClick` handler (used in modal)
- Separates `<LoginPage>` and `<LoginModal>` for different content and styling
- Adds `<ModalIntro>` component for use within modals. Modal content always passed as a child


src/components/LoginModal/login-modal.scss
```diff
@@ -0,0 +1,5 @@
+@import 'globals';
+
+.info {
```
sigmarstern: Shouldn't this be `.root`?

src/components/ModalWindow/index.js
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
+import styles from './modal-window.scss';
+import settings from 'settings';
+import CloseButton from 'components/CloseButton';
+
+const Modal = React.createClass({
```
sigmarstern: I'm not sure if this thing isn't just too smart for a dumb component. Shouldn't some logic be extracted and moved to [`ModalWindowContainer`](https://github.com/iris-dni/iris-frontend/blob/feature/181-signin-overlay/src/containers/ModalWindow.js)?

mattberridge: Perhaps, but we are triggering `this.props.hideModalWindow();` so I thought it was good here.

Likewise we could just `dispatch(hideModalWindow())` in the container too and we wouldn't need the dispatch action in this component.

PR #46, opened 2016-08-31
[RFR] Sign a petition

sigmarstern:
Adds a support button to the petition page.

unde_suscipit_odio_temporibus_possimus____iris-frontend

If user is currently not logged in, the button triggers a modal requesting the user to sign in through SSO. (Note the modal's custom intro text!)

unde_suscipit_odio_temporibus_possimus____iris-frontend

The modal is passing the user's intent (to support the petition) through the SSO flow. So the petition will be signed immediately if we're returning from SSO.


mattberridge:
Seems fine for me! Good stuff.

sigmarstern:
Ping @mattberridge and @Tom-Bonnike

PR #48, opened 2016-08-31
[RFR] City autocomplete

Tom-Bonnike:


src/actions/AutocompleteActions.js
```diff
@@ -0,0 +1,39 @@
...
+    // Depending on the endpoint, we can choose the repository we must search
+    // into.
+    switch (endpoint) {
+      case 'cities':
+        repository = citiesRepository;
+        break;
+      default:
+        console.warn(`No repository found for endpoint “${endpoint}“)`);
+        return;
+    }
```
Tom-Bonnike: Maybe there could be a list of all repositories and their names instead of having to manually add them here?

src/components/AutocompleteField/autocomplete-field.scss
```diff
@@ -0,0 +1,48 @@
...
+  border-color: color('grey');
+}
+
+.valid:focus {
+  border-color: color('dimmed');
+}
+
+.invalid {
+  border-color: color('error');
+}
```
Tom-Bonnike: A lot of repetition from the TextField component, but I don‘t think I have the choice?

src/components/Autocomplete/autocomplete.scss
```diff
@@ -0,0 +1,67 @@
...
+
+  &:last-of-type {
+    border-bottom: 0;
+  }
+}
+
+.suggestion > a {
+  text-decoration: none;
+  color: color('dimmed');
+}
```
Tom-Bonnike: Color should be directly on suggestion, that used to be for react-typeahead.

src/components/Autocomplete/autocomplete.scss
```diff
@@ -0,0 +1,67 @@
...
+  width: 50%;
+  list-style: none;
+}
+
+.focused,
+.suggestion:hover {
+  background-color: color('success');
+  border-color: color('success');
+  color: color('white');
+}
```
Tom-Bonnike: This should probably be after .suggestion declaration

src/components/AutocompleteField/index.js
```diff
@@ -0,0 +1,46 @@
...
+            <FormValidationIcon
+              error={fieldIsInvalid(helper)}
+              valid={fieldIsValid(helper)}
+            />
+          </div>
+        </div>
+      }
+    </div>
+  </div>
+);
```
Tom-Bonnike: Again, a lot of repetition from the TextField component, but couldn‘t find a good way to not repeat it.

The bottom validation part could be extracted as another component, though.

src/components/Autocomplete/index.js
```diff
@@ -0,0 +1,74 @@
...
+  if (query) {
+    query = query.trim().toLowerCase();
+
+    if (query.length > MIN_QUERY_LENGTH) {
+      return typeaheadSearch(endpoint, query);
+    }
+  }
+
+  return clearSuggestions();
+};
```
Tom-Bonnike: Should this method be in the AutocompleteContainer instead?

src/containers/Autocomplete.js
```diff
@@ -0,0 +1,35 @@
...
+
+const QUERY_THROTTLE_TIME = 700;
+
+const AutocompleteContainer = (props) => {
+  return <Autocomplete {...props} />;
+};
+
+AutocompleteContainer.propTypes = {
+  suggestions: React.PropTypes.array
+};
```
Tom-Bonnike: Should there be more propTypes declared?

src/components/Autocomplete/index.js
```diff
@@ -0,0 +1,115 @@
...
+  e.preventDefault();
+
+  // Update redux-form store manually to reflect the selection
+  helper.onChange({
+    data: suggestion,
+    id: suggestion.id,
+    class: 'City'
+  });
+};
+
```
Tom-Bonnike: Should some of these methods be in the AutocompleteContainer instead?

src/components/Autocomplete/index.js
```diff
@@ -0,0 +1,115 @@
...
+      autoComplete: html.autocomplete || 'off',
+      ...domOnlyProps(helper),
+      value: value || '',
+      onBlur: () => (handleBlur(value, suggestionFormatter, updateSuggestionInputValue, helper)),
+      onChange: (e, { newValue }) => (
+        handleChange(newValue, updateSuggestionInputValue, helper)
+      )
+    }}
+  />
+);
```
Tom-Bonnike: Lots of props, maybe I should add linebreaks and group some.

src/components/Autocomplete/index.js
```diff
@@ -0,0 +1,115 @@
...
+    focusFirstSuggestion
+    inputProps={{
+      name: name,
+      id: name,
+      required: html.required,
+      placeholder: html.placeholder,
+      autoComplete: html.autocomplete || 'off',
+      ...domOnlyProps(helper),
+      value: value || '',
+      onBlur: () => (handleBlur(value, suggestionFormatter, updateSuggestionInputValue, helper)),
```
Tom-Bonnike: Having to pass all of these reference to methods looks wrong, but is it?

src/containers/Autocomplete.js
```diff
@@ -0,0 +1,44 @@
...
+const QUERY_THROTTLE_TIME = 700;
+
+const AutocompleteContainer = (props) => {
+  return <Autocomplete {...props} />;
+};
+
+AutocompleteContainer.propTypes = {
+  suggestions: React.PropTypes.array,
+  value: React.PropTypes.string
+};
```
Tom-Bonnike: Should there be more propTypes declared here?

src/components/Autocomplete/index.js
```diff
@@ -0,0 +1,124 @@
...
+      id: name,
+      required: html.required,
+      placeholder: html.placeholder,
+      autoComplete: html.autocomplete || 'off',
+
+      ...domOnlyProps(helper),
+
+      value: value,
+      onBlur: () => (
+        handleBlur(value, suggestionFormatter, updateSuggestionInputValue, helper)
```
Tom-Bonnike: Having to pass all of these reference to methods looks wrong, but is it?

package.json
```diff
@@ -73,6 +74,7 @@
     "style-loader": "^0.13.1",
     "svg4everybody": "^2.1.0",
     "svgo": "^0.6.6",
+    "throttle-debounce": "^1.0.1",
```
mattberridge: we generally use `{ throttle, debounce }` from 'lodash` - if you can replace it easily, for consistency?

src/components/Autocomplete/autocomplete.scss
```diff
@@ -0,0 +1,63 @@
...
+  border-color: color('error');
+}
+
+.list {
+  @include map-declarations($form-fields-defaults, $form-fields-custom);
+  padding: 0;
+  border-top: 0;
+  border-color: color('grey');
+  border-radius: 0 0 $default-border-radius $default-border-radius;
+  width: 50%;
```
mattberridge: I would try and use a wrapping component and always use `width: 100%` - get the wrapper define the width of the component

src/components/Autocomplete/autocomplete.scss
```diff
@@ -0,0 +1,63 @@
+@import 'globals';
+
+$default-border-radius: map-get($form-fields-defaults, 'border-radius');
+
+.input {
+  @include map-declarations($form-fields-defaults, $form-fields-custom);
+  display: block;
+  width: 50%;
```
mattberridge: I would try and use a wrapping component and always use `width: 100%` - get the wrapper define the width of the component

src/components/Autocomplete/autocomplete.scss
```diff
@@ -0,0 +1,63 @@
...
+  list-style: none;
+}
+
+.suggestion {
+  @include map-declarations($form-fields-defaults, $form-fields-custom);
+  border: 1px solid color('grey-lite');
+  border-left: 0;
+  border-right: 0;
+  border-radius: 0;
```
mattberridge: is this needed?

src/components/Autocomplete/autocomplete.scss
```diff
@@ -0,0 +1,63 @@
...
+  border-radius: 0 0 $default-border-radius $default-border-radius;
+  width: 50%;
+  list-style: none;
+}
+
+.suggestion {
+  @include map-declarations($form-fields-defaults, $form-fields-custom);
+  border: 1px solid color('grey-lite');
+  border-left: 0;
+  border-right: 0;
```
mattberridge: i would just set explicit `border-top` and `border-bottom` - rather than undoing

src/containers/Autocomplete.js
mattberridge: yes, whatever is in `{...props}` :)

src/reducers/autocomplete.js
```diff
@@ -0,0 +1,29 @@
...
+        suggestions: []
+      });
+    case UPDATE_SUGGESTION_INPUT_VALUE:
+      return Object.assign({}, state, {
+        value: action.value
+      });
+    default:
+      return state;
+  }
+}
```
mattberridge: think you could try and write some tests for these?

src/components/AutocompleteField/index.js
mattberridge: Hmm, I'm not sure what to suggest. Could we make this a wrapping / higher order component that accepts a child and share it between input and autocomplete?

So

```
<div className={styles.wrapper}>
    {this.props.children}
```

instead of

```
<div className={styles.wrapper}>
  <Autocomplete
    helper={helper}
    {...config}
  />
```

mattberridge:
You've nailed this Tom, it looks awesome! Super impressed.
Added some comments to consider.

We need to fix the tests and then we can merge it for now I think and come back to the rest? Up to you...

src/components/Autocomplete/index.js
mattberridge: It's not wrong but we could change the component to a regular component with ES6 `class` or `React.createClass` - then reference `this.props` instead.

src/components/AutocompleteField/index.js
Tom-Bonnike: Seems like a good plan. I‘ll try

src/components/Autocomplete/index.js
Tom-Bonnike: Yes, I guess that it would probably look better with React.createClass. I‘ll try

src/components/Autocomplete/index.js
```diff
@@ -0,0 +1,117 @@
...
+          onChange: (e, { newValue }) => (
+            this.handleChange(newValue, {...this.props})
+          )
+        }}
+      />
+    );
+  }
+});
+
+export default Autocomplete;
```
Tom-Bonnike: @mattberridge Looks a bit better... I guess 🌵

src/components/FormFieldWrapper/index.js
```diff
@@ -14,6 +15,33 @@ const getClassname = (element, error) => {
...
+          id={config.name}
+          // pass props from config e.g. type, placeholder, maxlength
+          {...config.html}
+          // domOnlyProps required with latest react and redux-form 5.x
+          // see: https://github.com/erikras/redux-form/issues/1441#issuecomment-236966387
+          {...domOnlyProps(helper)}
+        />
+      );
+  }
+};
```
Tom-Bonnike: @mattberridge 
I don‘t think I could‘ve done this differently?
If we ever add input types (like the image picker), we‘ll simply add the necessary wrappers and actual element in the switch.

src/components/Autocomplete/index.js
```diff
@@ -0,0 +1,122 @@
...
+
+  handleSelection (e, { suggestion }, helper) {
+    // Disable form submitting when selecting option with ENTER
+    e.preventDefault();
+
+    // Update redux-form store manually to reflect the selection
+    helper.onChange({
+      data: suggestion,
+      id: suggestion.id,
+      class: 'City'
```
mattberridge: this needs to be dynamic. I am not sure how we best pass a data shape - maybe a function as a prop?

Tom-Bonnike: Ah yeah true!

src/actions/AutocompleteActions.js
```diff
@@ -5,33 +5,36 @@ import {
...
+export function typeaheadSearch (endpoint, query, id) {
   return (dispatch, getState) => {
     const repository = Repositories[endpoint];
 
     if (repository) {
       return repository.search(query).then(response => dispatch(
-        updateSuggestions(response.data)
+        updateSuggestions(response.data, id)
       ));
     }
```
Tom-Bonnike: We pass an identifier at each action. Here, it is the name of the autocomplete.

src/components/Autocomplete/index.js
```diff
@@ -82,7 +82,9 @@ const Autocomplete = React.createClass({
         focusFirstSuggestion
 
         onSuggestionsFetchRequested={(e) => (this.handleFetchRequest(e))}
-        onSuggestionsClearRequested={this.props.clearSuggestions}
+        onSuggestionsClearRequested={() => (
+          this.props.clearSuggestions(this.props.name)
+        )}
         onSuggestionSelected={(e, suggestion) => (
           this.handleSelection(e, suggestion)
         )}
```
Tom-Bonnike: We get the identifier from the props.

src/containers/Autocomplete.js
```diff
@@ -28,21 +28,25 @@ AutocompleteContainer.propTypes = {
...
-  value: autocomplete.value
-});
+export const mapStateToProps = ({ autocomplete }, ownProps) => {
+  const instance = autocomplete.instances[ownProps.name];
+
+  return {
+    suggestions: (instance && instance.suggestions) || [],
+    value: (instance && instance.value) || ''
+  };
+};
```
Tom-Bonnike: When mapping the state, we simply pick the right instance to get the data from.

src/reducers/autocomplete.js
```diff
@@ -5,24 +5,44 @@ import {
...
+      instances[action.id].value = action.value;
+
       return Object.assign({}, state, {
-        value: action.value
+        instances: instances
       });
+
     default:
       return state;
   }
```
Tom-Bonnike: And here, we pick the right instance to update its values.

PR #52, opened 2016-09-02
[RFR] Refactor PetitionContainer to have all petition data in place

sigmarstern:
@mattberridge As we said..

### DIFF
--- a/src/containers/Petition.js
+++ b/src/containers/Petition.js
@@ -9,26 +9,30 @@ const PetitionContainer = React.createClass({
   // When the component gets added to the DOM,
   // fetch Petition if `id` is not defined.
   componentWillMount () {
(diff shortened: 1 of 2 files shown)


src/containers/Petition.js
```diff
@@ -38,7 +42,7 @@ PetitionContainer.fetchData = ({ store, params }) => {
   return store.dispatch(fetchPetition(params.id));
 };
 
-export const mapStateToProps = ({ petition }) => getPetition(petition);
+export const mapStateToProps = ({ petition }) => ({petition: getPetition(petition)});
```
mattberridge: `({ petition: getPetition(petition) });` for syntax

sigmarstern: You're right. Since this is a pretty long line, I'll wrap and indent it

PR #56, opened 2016-09-07
Fix petitions sorting and only fetch non-draft petitions by default

artibella:
Petitions are sorted by created descending by default (might change in the future)

Also the default call now excludes draft and pending petitions from being listed.


artibella:
Fix was applied to different branch. Closed

PR #61, opened 2016-09-08
Feature/fix supportable

sigmarstern:
- Refactored three petition state related selectors for better reusability.
- Do not check for `timeMetric.votingActive` when rendering the support button, since all our newly created petitions are currently returning `supporters.required = 0`.

@mattberridge, @Tom-Bonnike, @artibella: Can we just keep the current check for supportability until the API returns proper values here?


mattberridge:
Great!

PR #62, opened 2016-09-08
[RFR] Sprint 4 Fixes

mattberridge:
Bug fixes we found in the review plus some other improvements:

- Tweak client-side loading of Petition and Petitions
- Add loading screen and (`<Loading>` component) to Petition and Petitions when redux is performing actions

### DIFF
--- a/src/containers/Petitions.js
+++ b/src/containers/Petitions.js
@@ -7,14 +7,8 @@ import Petitions from 'components/Petitions';
 import getPetitions from 'selectors/petitions';
 
 const PetitionsContainer = React.createClass({
-  // When the component gets added to the DOM,
-  // fetch Petitions if none are defined (clientside)
-  componentWillMount () {
-    const { petitions } = this.props;
-
-    if (!petitions.length) {
-      this.props.fetchPetitions(this.props);
-    }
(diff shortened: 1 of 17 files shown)


mattberridge:
Ping @sigmarstern for review

src/containers/Petitions.js
```diff
@@ -7,8 +7,14 @@ import Petitions from 'components/Petitions';
...
-  componentDidMount () {
-    this.props.fetchPetitions(this.props);
+  // When the component gets added to the DOM,
+  // fetch Petitions if none are defined (clientside)
+  componentWillMount () {
+    const { petitions } = this.props;
+
+    if (!petitions.length) {
+      this.props.fetchPetitions(this.props);
+    }
```
sigmarstern: This will break client side pagination. First page's petitions won't be replaced by those from second page. I guess we're good to always fetch petitions here.

mattberridge: hmm yeah true. but it always re-renders on the client on first load which is not ideal.
can we find a way to determine if the params / filters have changed when we do the filtering story?

sigmarstern:
👍

PR #63, opened 2016-09-08
Feature/allow nested api params

sigmarstern:
- removed `query-params` package in favor of our own [`encodeParams`](https://github.com/iris-dni/iris-frontend/blob/695972b65baa60b65aee50af2e5e0d65fdcf1564/src/helpers/encodeParams.js) helper which knows how to handle nested array/object params
- found out that the API accepts comma separated values for array params only. So, while `state=supportable.active,supportable.winner` would work, `state[]=supportable.active&state[]=supportable.winner` wouldn't
- thus fixed `getPetitionsRequestParams` helper to return the `state` param in the proper comma separated form

Ping @mattberridge

### DIFF
--- a/src/helpers/encodeParams.js
+++ b/src/helpers/encodeParams.js
@@ -1,0 +1,30 @@
+const serializeParam = (prefix, param, value) => {
+  const prefixedParam = prefix ? `${prefix}[${param}]` : param;
+
+  if (typeof (value) === 'object') {
+    return Object.keys(value)
+      .map(key => serializeParam(prefixedParam, parseKey(key), value[key]))
+      .join('&');
+  } else {
+    return [prefixedParam, value].join('=');
+  }
+};
+
+const parseKey = (key) => {
+  return parseInt(key) >= 0
+    ? ''
+    : key;
+};
+
+export default (params) => {
+  switch (params) {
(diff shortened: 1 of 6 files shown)


src/helpers/encodeParams.js
```diff
@@ -0,0 +1,30 @@
+const serializeParam = (prefix, param, value) => {
+  prefix = prefix ? `${prefix}[${param}]` : param;
+
+  if (typeof (value) === 'object') {
+    return Object.keys(value)
+      .map(key => serializeParam(prefix, parseKey(key), value[key]))
+      .join('&');
+  } else {
+    return `${prefix}=${value}`;
```
mattberridge: `[prefix, value].join('=')` ?

src/helpers/encodeParams.js
```diff
@@ -0,0 +1,30 @@
+const serializeParam = (prefix, param, value) => {
+  prefix = prefix ? `${prefix}[${param}]` : param;
```
mattberridge: it's not common to over-write JS function params like this sadly.
think you have to re-name the variable

sigmarstern: ok

PR #66, opened 2016-09-09
[WIP] Feature/filter petitions by city

sigmarstern:


src/containers/Petitions.js
```diff
@@ -1,26 +1,76 @@
...
+const PetitionsContainer = withRouter(React.createClass({
   componentWillMount () {
     this.props.fetchPetitions(this.props);
   },
 
+  componentWillReceiveProps (nextProps) {
+    if (this.props.location.pathname !== nextProps.location.pathname) {
+      this.props.fetchPetitions(nextProps);
+    }
+  },
```
Tom-Bonnike: I don‘t know if there‘s a better way to do this?
I want to update the petitions only when the location has changed (on selection of a suggestion in the input).

src/components/Petitions/index.js
```diff
@@ -1,38 +1,43 @@
...
+              return (
+                <GridItem key={petition.id}>
+                  <Teaser {...petition} key={petition.id} />
+                </GridItem>
+              );
+            })}
+          </Grid>
+          : <TextCenter>
+            There are no petitions for this city yet! <Link to='/petitions/new'>Create the first!</Link>
+          </TextCenter>
```
Tom-Bonnike: Probably will need some more styling.

src/components/Petitions/index.js
```diff
@@ -1,38 +1,43 @@
...
         <PageTitle
           title={settings.petitionsPage.title}
           centered>
-          <ButtonLink
-            href={'/petitions/new'}
-            text={settings.petitionsPage.createButton.text}
-            modifier={'accent'}
-          />
+
+          <Autocomplete {...autocompleteProps} />
```
Tom-Bonnike: Will probably be moved outside of the page title for more styling or to put it in a PetitionFilters component.

src/components/Petitions/index.js
mattberridge: I've done the fallback etc. logic in the pagination branch, but we will need to align.

src/containers/Petitions.js
mattberridge: I am working on this in the pagination branch. The idea is, all params / query params are stored in the state e.g.

```
{
  city: 'xx',
  page: 2,
  limit: 50
}
```

Then we deep compare whether any params have changed, then call `this.props.fetchPetitions` if so

src/containers/Petitions.js
```diff
@@ -1,26 +1,76 @@
...
+          } else {
+            router.push(petitionsPath());
+          }
+        },
+        onBlur () {}
+      },
+      html: {
+        placeholder: 'Filter by city'
+      }
+    };
```
pgoetze: This looks like there’s a lot of stuff going on in the return. Any chance to move a bit of the logic into helpers (e.g. the `suggestionFormatter` and `onChange`)?

mattberridge: `onChange` would we not dispatch the `fetchPetitions` action with new params, based on the city? this would do a client-side fetch

src/reducers/petitions.js
```diff
@@ -15,6 +16,10 @@ export default function petitions (state = initialState, action) {
       return Object.assign({}, state,
         action.petitions, { isLoading: false }
       );
+    case UPDATE_CITY_FILTER_VALUE:
```
mattberridge: For consistency; I think we should just use the `fetchPetitions action` and re-use `REQUEST_PETITIONS` and `RECEIVE_PETITIONS` actions. This is what will happen in the pagination

Tom-Bonnike: I don‘t get what you mean? This action is only there to update the input value with the chosen suggestion from the search input.
Or are you saying this for when we want to prefill the input / change the headline?

src/containers/Petitions.js
Tom-Bonnike: This is already what‘s being done.
The location changes, but this doesn‘t automatically trigger a new fetchPetitions, as `componentWillMount` is not called again. If I directly make the `fetchPetitions` call in the onChange, the `props.location` which is used in it is still the old value, which would display the same petitions again.
That‘s why I put it in `componentWillReceiveProps` and do the check to see if the location changed there. I don‘t think there was a more suited event.

src/containers/Petitions.js
```diff
@@ -1,26 +1,71 @@
...
+          data: cityFilterValue
+        },
+        onChange (newValue) {
+          updateCityFilterValue(newValue);
+
+          if (newValue.id) {
+            router.push(petitionsPath({city: newValue}));
+          } else {
+            router.push(petitionsPath());
+          }
```
mattberridge: cleaner?

```
          router.push(petitionsPath({
            city: newValue && newValue.id ? newValue : ''
          }));
```

src/containers/Petitions.js
```diff
@@ -1,51 +1,97 @@
...
+  }),
+
+  getTitle () {
+    return this.props.city
+      ? `${settings.petitionsPage.titleLocalized} ${this.props.city.name}`
+      : settings.petitionsPage.title;
+  },
+
   render () {
+    const title = this.getTitle();
```
mattberridge: stuff like this always seems like a smell to me. Can't we know the `title` already when we render?

mattberridge: I think we can move `getTitle` to `mapStateToProps` and just pass a prop called `title`.
Use a helper like `getPageTitle(city)`

Tom-Bonnike: Ah yeah, better. 👍

src/reducers/petitions.js
```diff
@@ -20,6 +22,14 @@ export default function petitions (state = initialState, action) {
           qs: action.qs || ''
         }
       );
+    case GET_CITY:
```
mattberridge: Genuine question - why do we need two `city` state items?
Isn't the `city` already defined in `cityFilterValue.data` ?

Or have I missed something?

Tom-Bonnike: Oh fuck. You‘re right.

src/helpers/encodeParams.js
```diff
@@ -4,9 +4,10 @@ const serializeParam = (prefix, param, value) => {
   if (typeof (value) === 'object') {
     return Object.keys(value)
       .map(key => serializeParam(prefixedParam, parseKey(key), value[key]))
+      .filter(serializedParam => !!serializedParam)
       .join('&');
   } else {
-    return [prefixedParam, value].join('=');
+    return value && [prefixedParam, value].join('=');
```
pgoetze: Just because I’m seeing a couple of different approaches in the project: What is the preferred way to handle such string concatenations? Using `join`, or using string interpolation like \`${prefixed}=${value}\`? I guess it’s not `+`, right?

mattberridge: To tell you the truth it differs for everyone :)

If it's a consistent separator I will always use `join()`

```
['query', param].join('=')
[param1, param2].join(',')
['path', id 'otherpath'].join('/')
```

Otherwise, when there are different separators use `${}`

```
`http://${domain}/path/path?${query}`
```

src/components/PetitionsFilters/petitions-filters.scss
```diff
@@ -0,0 +1,47 @@
...
+  z-index: 2;
+
+  @include media('>460px') {
+    display: block;
+  }
+}
+
+.input-wrapper {
+  display: inline-block;
+  width: 240px;
```
mattberridge: wrap this in a `max-width` query

src/components/PetitionsFilters/petitions-filters.scss
```diff
@@ -0,0 +1,47 @@
...
+
+.label {
+  display: inline-block;
+  color: color('black');
+  font-weight: bold;
+  cursor: pointer;
+}
+
+.icon {
+  display: none;
```
mattberridge: wrap this in a `max-width` query

src/helpers/encodeParams.js
pgoetze: Cool, thanks @mattberridge, good to know.

src/components/PetitionsFilters/petitions-filters.scss
```diff
@@ -0,0 +1,47 @@
...
+  position: absolute;
+  right: 1em;
+  top: 50%;
+  transform: translateY(-50%);
+  z-index: 2;
+
+  @include media('>460px') {
```
mattberridge: can be removed with the above

mattberridge:
Awesome @Tom-Bonnike - a few comments but kudos for getting the autocomplete in there so easily (and making it so re-useable in the first place!)

src/helpers/getPaginationLink.js
```diff
@@ -1,3 +1,11 @@
-export default (page, qs) => {
-  return `/petitions/page/${page}${qs ? `?${qs}` : ''}`;
+export default (params, page, qs) => {
+  var petitionsPath = '/petitions/';
+
+  if (params.cityName && params.city) {
+    petitionsPath += `${params.cityName}-${params.city}/`;
+  }
+
+  petitionsPath += `page/${page}${qs ? `?${qs}` : ''}`;
```
mattberridge: won't this always return, missing `else` ?

mattberridge: forget it! ;)

Tom-Bonnike: haha, why?

src/selectors/petitionCity.js
```diff
@@ -7,8 +7,5 @@ const cityLabel = (city = {}) => {
 export default (petition = {}) => {
   const city = petition.city && petition.city.data;
 
-  return {
-    ...city,
-    label: cityLabel(city)
-  };
+  return city && { ...city, label: cityLabel(city) };
```
Tom-Bonnike: … Does any of you know why I suddenly have to make that check? :p

src/services/api/repositories/petition.js
```diff
@@ -23,6 +23,11 @@ export default {
   update: (petition) => {
     const requestPath = path.join('/petitions', petition.id.toString());
     delete petition.id;
+
+    if (petition.city && !petition.city.data) {
+      delete petition.city.data;
+    }
```
Tom-Bonnike: And this one? Maybe this was an API change… If you pass a null object as the data property to the city on a form you‘re updating, the request fails as you need to pass an empty object/nothing but not null.

src/components/Petitions/index.js
```diff
@@ -1,30 +1,45 @@
...
           petitions={petitions}
           isLoading={isLoading}
           />
         : <NoResults />
       }
-      {petitions.length > 0 &&
-        <Pagination />
+
+      {petitions.length > 0 && totalCount > petitions.length &&
+        <Pagination params={params} />
```
mattberridge: can we get `params` in `containers/Pagination.js` ?

Tom-Bonnike: Yeah, but I will have to set the Pagination container as a “withRouter“ class then

src/components/Petitions/index.js
```diff
@@ -1,30 +1,45 @@
...
         ? <TeaserGrid
           petitions={petitions}
           isLoading={isLoading}
           />
         : <NoResults />
       }
-      {petitions.length > 0 &&
-        <Pagination />
+
+      {petitions.length > 0 && totalCount > petitions.length &&
```
mattberridge: seems a bit messy here? as well as @eschaefer added a check inside `Pagination` to check for rendering.

if it's absolutely crucial, add a `showPagination` boolean calculated in the container?

eschaefer: Ja, I am letting `<Pagination />` decide for itself if it should render. No need to check for length here.

Tom-Bonnike: ? So the pagination isn‘t supposed to be rendered when there‘s only one page? That‘s how it was

src/components/Petitions/index.js
mattberridge: I had planned to keep `params` in the `petitions` object in the state. can we add `cityName` and not use the router?

src/containers/Petitions.js
```diff
@@ -1,51 +1,91 @@
...
+
   render () {
     return (
       <div>
-        <Helmet title={settings.petitionsPage.title} />
-        <Petitions {...this.props} />
+        <Helmet title={this.props.title} />
+        <Petitions
+          {...this.props}
+          title={this.props.title}
```
mattberridge: not needed, as you're already spreading all the props in the line above? 😉

Tom-Bonnike: ah yeah, just stupidly copied it from the helmet :)

src/helpers/getPaginationLink.js
```diff
@@ -1,3 +1,11 @@
...
+
+  return petitionsPath;
 };
```
mattberridge: I've just realised we have come this far without ever using the `path` helper from node, to generate paths 😬

src/reducers/petitions.js
```diff
@@ -16,10 +17,15 @@ export default function petitions (state = initialState, action) {
       return Object.assign({}, state,
         action.petitions, {
           isLoading: false,
+          totalCount: action.petitions.total,
```
mattberridge: you might remove this if we don't need to know it until inside the `Pagination` component? (which has its own helper to generate props)

src/helpers/getPaginationLink.js
mattberridge: sorry... I added a bullshit comment not understanding the code, and then deleted it!

src/services/api/repositories/petition.js
mattberridge: yeah this seems strange. please clarify with Lovely? they changed @pgoetze's request yesterday so that might be what broke it?

src/containers/Petitions.js
```diff
@@ -1,51 +1,90 @@
...
 
 PetitionsContainer.propTypes = {
   petitions: React.PropTypes.array,
-  fetchPetitions: React.PropTypes.func
+  total: React.PropTypes.number,
+  isLoading: React.PropTypes.bool,
+  title: React.PropTypes.string,
+  currentCity: React.PropTypes.object,
+  fetchPetitions: React.PropTypes.func,
+  citySuggestionFormatter: React.PropTypes.func
```
mattberridge: can't see this used from props anywhere?

src/helpers/getPetitionsPageTitle.js
```diff
@@ -0,0 +1,9 @@
+import settings from 'settings';
+
+export default (currentCity) => {
+  const city = currentCity && currentCity.name;
+
+  return city
+    ? `${settings.petitionsPage.titleLocalized} ${city}`
+    : settings.petitionsPage.title;
+};
```
pgoetze: This piece of code should get a test as well. But since there is a bit of time pressure and it is working for now, we can add them in a later PR I guess. I was working on a couple of helper tests anyway on a separate branch.

PR #67, opened 2016-09-09
[RFR] Pagination

mattberridge:


src/routes.js
```diff
@@ -27,6 +27,7 @@ export default function (props = {}) {
         <Route path='auth/login' component={LoginPage} />
         <Route path='auth/logout' component={Logout} />
         <Route path='petitions' component={Petitions} />
+        <Route path='petitions/page/:page' component={Petitions} />
```
sigmarstern: Just found out you can make the page param optional: `<Route path='petitions(/page/:page)' component={Petitions} />` and have only one route definition.

Will be helpfull when adding more params like `city` and `state`.

mattberridge: Thanks!

artibella:
@eschaefer: On mobile there is now not enough room for the page display



Otherwise looking good 👍

test/helpers/getPaginationLink.js
```diff
@@ -0,0 +1,20 @@
+import chai from 'chai';
+import getPaginationLink from 'helpers/getPaginationLink';
+
+const { assert } = chai;
```
pgoetze: Should we use 

```
import { assert } from 'chai';
```

instead of 
```
import chai from 'chai';
const { assert } = chai;
```
? Same here https://github.com/iris-dni/iris-frontend/pull/67/files#diff-65b5fd2ae5e0bcfe4b9d044f60ee077eR5

eschaefer:
@artibella I just added mobile size styles

test/helpers/getPaginationLink.js
eschaefer: Yep, I'd say it's cleaner.

PR #70, opened 2016-09-12
use /petitions view as the default homepage

eschaefer:
Visit https://iris-frontend-staging-pr-70.herokuapp.com, see list of petitions as home page.

Old "home" still at https://iris-frontend-staging-pr-70.herokuapp.com/home


mattberridge:
👍

PR #72, opened 2016-09-13
Ensure use of https

pgoetze:
It redirects all http requests to https.
You need to use https urls for the `BASE_URL` and `API_URL` ENV variables.


pgoetze:
@artibella Should we handle configured non-https ENV urls (not redirecting if the `BASE_URL`uses http), or should we just enforce the use of https in the urls in the ENV variables?

PR #74, opened 2016-09-13
Fix home link in navigation & default route redirect

artibella:


pgoetze:
Do we still need the Home view, now that it’s not reachable anymore? Else it could be removed.

PR #75, opened 2016-09-13
[RFR] Pagination unit test

eschaefer:
Enzyme has a straightforward shallow-render method to use w/ Chai. Might as well...

### DIFF
--- a/test/components/Pagination.js
+++ b/test/components/Pagination.js
@@ -1,40 +1,56 @@
 import React from 'react';
 import { Link } from 'react-router';
 import { shallow } from 'enzyme';
-import { expect } from 'chai';
+import { assert } from 'chai';
 import Pagination from 'components/Pagination';
 import styles from 'components/Pagination/pagination.scss';
 
 describe('<Pagination />', () => {
   it('does not render without > 1 pages', () => {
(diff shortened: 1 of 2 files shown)


test/components/Pagination.js
```diff
@@ -0,0 +1,40 @@
...
+    const wrapper = shallow(<Pagination totalPages={4} isFirstPage isLastPage={false} />);
+    expect(wrapper.find(Link)).to.have.length(1);
+    expect(wrapper.find(Link).props().rel).to.equal('next');
+  });
+
+  it('renders both "previous" and "next" links if not first or last page', () => {
+    const wrapper = shallow(<Pagination totalPages={4} isLastPage={false} isFirstPage={false} />);
+    expect(wrapper.find(Link)).to.have.length(2);
+  });
+});
```
pgoetze: I think we agreed on using `assert` instead of `expect` in the beginning and have the tests in form of 

```
const actual = ...;
const expected = ...;

assert.<whatever>(actual, expected);
```

Does it make sense to keep it consistent?

eschaefer: Ah ok, makes sense. Changing em now

PR #79, opened 2016-09-14
[RFR] Linting fixes for atom

eschaefer:
When I first set up the project, linting didn't work out of the box on Atom. 

Adding these lines so it works.


mattberridge:
Merge it!

PR #82, opened 2016-09-14
Replaced petition react-router link with our own stylized Link component

Tom-Bonnike:
@eschaefer => I don‘t really know if there‘s a downside to not using the react-router link, but we have our own Link component with style & hover state.
As long as we don‘t need to use the react-router‘s “activeClassName” property, I don‘t think it‘s that bad to not use it?

### DIFF
--- a/src/components/PetitionInfo/index.js
+++ b/src/components/PetitionInfo/index.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import styles from './petition-info.scss';
 import IconAndInfo from 'components/IconAndInfo';
-import { Link } from 'react-router';
+import Link from 'components/Link';
 import { petitionsPath } from 'helpers/petitionUrls';
 
 const PetitionInfo = ({ owner, city, dateRange }) => (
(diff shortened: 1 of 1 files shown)


eschaefer:
Probably fine to use our own unless you need to do fancy router things (params, query, etc)

Tom-Bonnike:
Merging then :)

mattberridge:
If you don't use `{ Link } from 'react-router'` then you don't get client-side routing at all?

Therefore when you click the link you just changed, you're re-rendering the whole page (ofc), but it _could_ be handled client-side just by requesting `fetchPetitions` to get the new Petitions into the content.

Whether this app needs to be / have full client-side rendering is a different topic, just letting you know (if you didn't!)

Tom-Bonnike:
Ahhhh, makes sense yes. I’ll probably revert that change then…

We could either:
- use @eschaefer solution and apply the styles we have on the Link component to all the `a` tags
- rename our Link component to LinkWrapper or whatever, and use the react-router Link in it…

I don’t know. What would be best in your opinion? I just thought the link to the city seemed a bit dull compared to the Link component.

PR #83, opened 2016-09-14
[RFR] Client-side Loading Changes

mattberridge:
- Removes FOUC when loading a petition client-side in place of loading screen (to be styled)
- Deep compares route and state params now, to see whether Petitions filters have changed, before fetching petitions


mattberridge:
Ping @Tom-Bonnike @pgoetze @eschaefer for review.

Sure this might change as still stabbing in the dark a bit, but it seems to sort out a couple of the issues above

PR #84, opened 2016-09-14
Allow setting an SSO provider via env variables

pgoetze:
We need the provider to be configurable for our different environments (Dev should use the dev SSO, staging should use the staging SSO, etc.)

The provider configured in the env vars and the ones from `config.js` are merged.
Only providers with non-empty text and url are used.

The env vars for setting the SSO provider are:
- `SSO_PROVIDER_TEXT`
- `SSO_PROVIDER_URL`

ping @eschaefer, @mattberridge, @Tom-Bonnike for review

(I have no idea how to test this helper with all the internal stuff being loaded. I guess this means it’s poorly designed 😕 , any ideas how to improve it?)


mattberridge:
can you define test `ENV` values in `webpack.config.test.js`?

mattberridge:
but yeah it's not that testable. unless you pass `settings` in as an argument to the function to make it closer to being 'pure'?

pgoetze:
Jep, defining test `ENV` values should help to test it against the env provider. I would like passing the ssoProviders as a param! I think I will change it.

pgoetze:
@mattberridge ssoProviders are now passed as argument.

test/helpers/generateSsoProviders.js
```diff
@@ -7,15 +7,14 @@ describe('generateSsoProviders', () => {
   const returnPath = '/return-url';
   const escapedReturnUrl = escape([envVars.baseUrl, returnPath].join(''));
 
+  const ssoProvidersFromEnv = envVars.ssoProviderText && envVars.ssoProviderUrl
```
mattberridge: @pgoetze I added this so tests won't fail when you haven't defined any providers in `ENV`.

PR #86, opened 2016-09-15
[RFR] Loading styles update

eschaefer:
Adds a CircleProgressBar with an interval percentage change, to mimic a spinner.

untitled

- [x] Use this on the `<TeaserGrid>`


src/components/Loading/index.js
```diff
@@ -1,9 +1,43 @@
...
-    : children
-);
+const Loading = React.createClass({
+
+  getInitialState () {
+    return { percentage: 75 };
+  },
+
+  tick () {
+    this.setState({ percentage: this.state.percentage + 125 });
```
eschaefer: Ok I didn't see you using `setState` anywhere else, but seems like overkill to have a redux action for this :/

Tom-Bonnike: I had the same problem, I just had to update a simple text input with a value…
I still made redux actions for those, but you’re right, it sucks :/

eschaefer: @mattberridge whaddya think? the only advantage to having it in redux is time travel / replay, right? Abromov suggests [it's nbd](https://twitter.com/dan_abramov/status/725089243836588032) in this kinda thing.

src/components/Loading/index.js
```diff
@@ -1,9 +1,43 @@
...
+  },
+
+  render () {
+    const { isLoading, children, onServer } = this.props;
+
+    if (onServer || (__CLIENT__ && isLoading) || !children) {
+      return (
+        <div
+          className={styles.root}
+          aria-busy='true'>
```
mattberridge: 👍

src/components/Loading/index.js
```diff
@@ -1,9 +1,43 @@
...
-const Loading = ({ isLoading, children, onServer }) => (
-  onServer || (__CLIENT__ && isLoading) || !children
-    ? <div>Loading screen</div>
-    : children
-);
+const Loading = React.createClass({
+
+  getInitialState () {
+    return { percentage: 75 };
+  },
```
mattberridge: No padding line +

```
  getInitialState: () => ({
    percentage: 75
  }),
```

src/components/Loading/index.js
mattberridge: I think you're right Redux is a bit overkill for this, unless we are measuring the actual progress of the API requests by bytes complete or so 😄 

I might have just gone for a CSS-only solution and I don't quite get the metrics of the initial percentage being `75` and adding `125` - can you clarify? But seems alright to me overall!

src/components/Loading/loading.scss
```diff
@@ -0,0 +1,8 @@
+@import 'globals';
+
+.root {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  height: 50vh;
```
mattberridge: this is weird because when you paginate you get a strange page jump whilst loading.

I'd set a `min-height` and perhaps something smaller?

src/components/Loading/loading.scss
```diff
@@ -0,0 +1,8 @@
+@import 'globals';
+
+.root {
+  display: flex;
```
mattberridge: any chance this might bork nested child components?

src/components/TeaserGrid/index.js
```diff
@@ -3,21 +3,24 @@ import styles from './teaser-grid.scss';
...
-      <Grid>
-        {(petitions || []).map((petition) => {
-          return (
-            <GridItem key={petition.id}>
-              <Teaser {...petition} key={petition.id} />
-            </GridItem>
-          );
-        })}
-      </Grid>
+  <Loading isLoading={isLoading}>
```
mattberridge: I could be up for keeping the `<Grid>` and the `<Loading>` components separate. What do you think?

mattberridge:
Going back to the question re: client-side routing and having a bit more thought: navigating the app feels a bit strange to me sometimes. Especially when things happen 'instantly'

Does anyone have an opinion on whether we should ditch client-side routing for some links, to give the user a bit more feedback re: routing and browser loading? And use it more for places like pagination / filtering.

Maybe we should meet and discuss?

@artibella

src/components/Loading/index.js
eschaefer: @mattberridge I'm all for a CSS-only solution instead. This was a band-aid since I saw using CircleProgressBar as a quick fix. 

Initial percentage of `75` then then adding `125` every second just staggers the animation based on what CircleProgressBar can already do. It's a hack tho.

Tom-Bonnike: Ah, there are two other places where I made very simple actions when a setState would have been enough. If you agree I‘ll update them as well then :)

mattberridge: @Tom-Bonnike what are the implementations?

I don't think we need change anything, especially if we switch to a CSS solution here. I'm happy with no `setState` at all, it's just that here the state is purely presentational

Tom-Bonnike: The three navigation actions and 2 I thought were a bit overkill (taking a look at the actions now, they’re used in multiple places so maybe it actually makes sense to create actions. :))
- Update current city (petitions actions)
- Update suggestion input value

Tom-Bonnike:
I think there is a bug on the animation spinning. To reproduce:
- Go to /petitions
- Set “throttle network” in inspector to some slow connection speed
- Filter by city

eschaefer:
@Tom-Bonnike yeah I see this bug too. Probably a good reason for me to just turn this into a pure CSS animation that doesn't depend on state to animate the spinner.

src/components/Loading/index.js
eschaefer: Killed it

eschaefer:
@Tom-Bonnike @mattberridge I quickly refactored a bit to not use `setState` and it also takes care of the bug on slow connections.

src/components/CircleProgressBar/index.js
```diff
@@ -15,11 +15,26 @@ const CircleProgressBar = React.createClass({
...
+
   componentDidMount () {
     this.initProgressBar();
     if (this.props.percentage > 0) {
       this.updateProgressBar(this.props.percentage);
     }
+
+    if (this.props.isSpinner) {
+      this.spinPercentage = 0.25;
+      this.interval = setInterval(this.infiniteSpin, 1000);
```
Tom-Bonnike: couldn’t you use RequestAnimationFrame here? With a much lower spinPercentage obviously.

eschaefer: Eh, won't gain much with that here. That's more useful when trying to do computationally intensive things inside of a short interval. Here I'm just incrementing a number once every second .

eschaefer:
Ge-mergeable? @mattberridge @Tom-Bonnike ?

PR #88, opened 2016-09-15
[RFR] Reworked custom Link component to allow react-router links to be used

Tom-Bonnike:
@mattberridge @pgoetze @eschaefer 

Solution so that Links always are a react-router link if they’re not external, and that styles are shared. Tell me what you think.

### DIFF
--- a/src/components/ButtonLink/index.js
+++ b/src/components/ButtonLink/index.js
@@ -1,21 +1,12 @@
 import React from 'react';
-import { Link } from 'react-router';
+import Link from 'components/Link';
 import { getClassname } from 'components/Button';
 
-const createAttrs = (props) => {
-  return {
-    [props.external ? 'href' : 'to']: props.href || '#',
-    onClick: props.onClick
-      ? (e) => { e.preventDefault(); props.onClick(); }
-      : () => {},
-    className: getClassname(props)
-  };
-};
-
 const ButtonLink = (props) => (
-  props.external
-    ? <a {...createAttrs(props)}>{props.children || props.text}</a>
-    : <Link {...createAttrs(props)}>{props.children || props.text}</Link>
+  <Link
+    className={getClassname(props)}
+    {...props}
+  />
 );
 
 export default ButtonLink;
--- a/src/components/Link/index.js
+++ b/src/components/Link/index.js
@@ -1,10 +1,23 @@
 import React from 'react';
+import { Link as RouterLink } from 'react-router';
 import styles from './link.scss';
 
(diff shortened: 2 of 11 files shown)


src/helpers/getPetitionLink.js
```diff
@@ -1,7 +1,11 @@
...
-    id
-  ].join('/');
+export default (id, fullURL) => {
+  let link = ['petitions', id];
+
+  if (fullURL) {
+    link.unshift(process.env.BASE_URL);
+  } else {
+    link[0] = `/${link[0]}`;
+  }
```
Tom-Bonnike: …

src/components/ButtonLink/index.js
```diff
@@ -1,21 +1,12 @@
...
-  props.external
-    ? <a {...createAttrs(props)}>{props.children || props.text}</a>
-    : <Link {...createAttrs(props)}>{props.children || props.text}</Link>
+  <Link
+    className={getClassname(props)}
+    {...props}
+  />
 );
 
 export default ButtonLink;
```
Tom-Bonnike: I guess we could maybe delete that component then…
I just moved the whole `createAttrs` logic to the Link component, see below :)

src/helpers/getPetitionLink.js
mattberridge: I would keep these two helpers separate for clarity I think

```
getPetitionPath
getPetitionURL
```

src/components/Link/index.js
```diff
@@ -1,10 +1,23 @@
...
+  props.external
+    ? <a {...createAttrs(props)}>
+      {props.children || props.text || props.href}
+    </a>
+    : <RouterLink {...createAttrs(props)}>
+      {props.children || props.text || props.href}
+    </RouterLink>
 );
 
 export default Link;
```
mattberridge: This is nice, works well!

src/components/ButtonLink/index.js
mattberridge: The thing with `<ButtonLink>` is that even though it only uses `<Link>` now it handles the class name construction. Things might be messy if we have to start using `import { getClassname } from 'components/Button'` further up the tree, on `<Link>` directly?

src/helpers/getPetitionLink.js
Tom-Bonnike: Oh, Github doesn’t hide the comments on old diffs with the new reviews, but this is done. Yes, good solution. :)

PR #90, opened 2016-09-16
[RFR] Sort by date & supporters amount

Tom-Bonnike:


src/actions/PetitionsActions.js
```diff
@@ -19,14 +19,15 @@ export function fetchPetitions ({ location, params }) {
     page: parseInt(params && params.page || query.page || 1),
     city: params && params.city || query.city || '',
     cityName: params && params.cityName || query.cityName || '',
-    limit: parseInt(query.limit || 12)
+    limit: parseInt(query.limit || 12),
+    sort: params && params.sort || query.sort || ''
```
Tom-Bonnike: oh, I guess I don’t need the tests for params. Sorting will only be done with query parameters

Tom-Bonnike: oh, I guess I don’t need the tests for params. Sorting will only be done with query parameters

src/containers/Petitions.js
```diff
@@ -22,43 +19,21 @@ const PetitionsContainer = withRouter(React.createClass({
   },
 
   componentWillReceiveProps (nextProps) {
-    if (this.props.location.pathname !== nextProps.location.pathname) {
+    if (this.props.location.pathname !== nextProps.location.pathname ||
+        this.props.location.search !== nextProps.location.search) {
```
Tom-Bonnike: @mattberridge 
Any better idea?

src/helpers/getPetitionsRequestParams.js
```diff
@@ -1,14 +1,20 @@
...
 
-export default ({ limit, page, city }) => {
+export default ({ limit, page, city, sort }) => {
   const saneLimit = sanitizeParamLimit(limit);
 
+  if (sort === 'supporters') {
+    sort = '-supporters.amount';
+  } else {
+    sort = '-created';
+  }
```
Tom-Bonnike: I guess for now we only need those two conditions, as the default is the same as “Most recent”

src/actions/PetitionsActions.js
mattberridge: just `query` :)

src/components/Petitions/index.js
```diff
@@ -8,35 +8,29 @@ import PetitionsFilters from 'components/PetitionsFilters';
 import NoResults from 'components/NoResults';
 import Pagination from 'containers/Pagination';
 
-const Petitions = ({
-  petitions,
-  isLoading,
-  title,
-  autocompleteProps
-}) => (
+const Petitions = (props) => (
```
mattberridge: why the style change?

Tom-Bonnike: … I‘ll revert that. There was probably a reason, when I did it :p

Tom-Bonnike: Ah… I remember now :) I need to pass all the props to the children PetitionsFilters component now, so I didn’t destructure the props object.

PR #91, opened 2016-09-16
[RFR] Status filtering

Tom-Bonnike:
Started from #90


src/actions/PetitionsActions.js
```diff
@@ -10,23 +10,34 @@ import {
...
 export function fetchPetitions ({ location, params }) {
-  // Get query from react-router locatiin
+  // Get query from react-router location
   const { query } = location;
 
-  // Construct our query params, based on
-  // route params or query string params
+  // Construct our query params based on route params or query string params
   const queryParams = {
     page: parseInt(params && params.page || query.page || 1),
```
eschaefer: Instead of always checking `params && params.something`, you can assign `params` a default value. 

```
export function fetchPetitions ({ location, params = {} ) { }
```

Makes for slighly DRY-er boolean checks.

src/actions/PetitionsActions.js
```diff
@@ -9,24 +9,35 @@ import {
...
-  // picking the relavent props for filering
+  // To avoid repetition between URL params and query params, we don’t save
+  // params taken from URL.
+  let encodedParams = [];
+
+  for (const key in queryParams) {
+    if (query[key]) {
+      encodedParams.push(key);
+    }
+  }
```
mattberridge: why can't we just use `pick` below here?
we are in control of what we want to use as real params vs query params? e.g. `['limit', 'sort']` are the query params?

src/components/PetitionsFilters/index.js
```diff
@@ -1,32 +1,147 @@
...
+          city: newValue.id ? newValue : '',
+          sort: location.query.sort || ''
+        }));
+      },
+      onBlur () {}
+    },
+    html: { placeholder: settings.petitionsPage.filters.city.placeholder }
+  }),
+
+  render () {
```
mattberridge: I think this component is a little big and complex. Any way we can break it down?

src/components/PetitionsFilters/index.js
```diff
@@ -1,32 +1,147 @@
...
+          </div>
+        </div>
+
+        <div className={styles.filter}>
+          <label className={styles.label} htmlFor={SORT_INPUT_NAME}>
+            {settings.petitionsPage.sortBy}
+          </label>
+
+          <div className={styles['input-wrapper']}>
+            <select
```
mattberridge: can we create a `<Select>` component?

src/helpers/getPaginationLink.js
```diff
@@ -1,6 +1,12 @@
 export default (params, page, qs) => {
   var petitionsPath = '/petitions/';
 
+  console.log(qs);
```
mattberridge: remove ;)

src/helpers/getPetitionsRequestParams.js
```diff
@@ -1,26 +1,45 @@
...
       'supportable.winner',
       'loser',
       'processing.*',
       'closed'
-    ].join(',')
-  };
-
-  if (city) {
-    params.city = city;
+    ].join();
```
mattberridge: this is getting complicated.
ask @sigmarstern if he has any ideas :)

src/components/PetitionsFilters/index.js
Tom-Bonnike: Ah yeah. Will do

src/helpers/getPetitionsRequestParams.js
Tom-Bonnike: Oh I guess I should have these arrays in some configuration file instead.

src/components/PetitionsFilters/index.js
Tom-Bonnike: Ah, yes.
FilterField w/ FilterLabel & FilterSelect / FilterInput (if needed) inside. Would already look much better. :)

Tom-Bonnike:
Ok, applied the (first?) changes you wanted @mattberridge. There’s still the routing bug now. We need to discuss it. :)

src/settings/config.js
```diff
@@ -161,5 +173,23 @@ export default {
   ],
   footer: {
     copyright: 'Copyright \u00A9 2016 Iris'
+  },
+  petitionsStates: {
```
mattberridge: Annoying but these shouldn't be in settings / over-writeable by the user. The need to be global constants elsewhere

PR #92, opened 2016-09-16
[RFR] Private/public fork setups for theming

eschaefer:


mattberridge:
I'd love to know what this is doing and how it works, if you can demo sometime :)

scripts/update.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+ORIGINAL_REPO=iris-dni/iris-frontend.git
+HAS_CONFIGURED_UPSTREAM=$(git config --get remote.upstream.url | grep $ORIGINAL_REPO)
+
+if [ "$HAS_CONFIGURED_UPSTREAM" ]; then
+  read -p "You should definitely stash any changes before doing this. Continue? [Y/n] " -n 1 -r
```
sigmarstern: Doesn't `[Y/n]` imply that `Y` is the default here and simply hitting return would do the update?

sigmarstern:
Temporarily close PR

package.json
```diff
@@ -8,7 +8,7 @@
     "clean": "rm -rf static/dist",
     "nodecheckversion": "node ./scripts/node-checkversion.js",
     "preinstall": "npm run nodecheckversion",
-    "postinstall": "sh ./scripts/upstream.sh && npm run build",
```
eschaefer: This only sets the upstream repo. Why remove? If this stays removed, please make a note for users on how to set their upstream.

mattberridge: We need to `run build` post install for deployment as well, I think

eschaefer: Ah ok. Can maybe just roll-up setting upstream into just the `update` script then.

sigmarstern: This would fail on Heroku where `npm install` is being executed in a `/tmp` folder which isn't a git repository.

eschaefer:
Okee, should be ready. None of this upstream/update stuff should run on Heroku automatically now.

PR #101, opened 2016-09-21
[RFR] Feature/check if petition was already supported by me

sigmarstern:
If queried with `?extend=supporting`, the API will return information if the current user has already supported a petition.

This PR adds a check for this property to disable the support button if necessary.

@Tom-Bonnike, @mattberridge

### DIFF
--- a/test/selectors/petition.js
+++ b/test/selectors/petition.js
@@ -1,0 +1,27 @@
+import { assert } from 'chai';
+import mockPetition from '../mocks/petition';
+import getPetition from 'selectors/petition';
+
+describe('getPetition', () => {
+  context('with a petition I have not supported yet', () => {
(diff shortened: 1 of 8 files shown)


test/selectors/petition.js
```diff
@@ -0,0 +1,28 @@
...
+import mockPetition from '../mocks/petition';
+import getPetition from 'selectors/petition';
+
+describe('getPetition', () => {
+  context('with a petition I have not supported yet', () => {
+    const petition = {
+      ...mockPetition.data,
+      extensions: { supporting: false }
+    };
+    console.log(petition);
```
mattberridge: remove

sigmarstern: Done.

PR #115, opened 2016-09-23
[RFR] Implement OG API

sigmarstern:
- Add `check(url)` to OG repository
- Add `fetchOg`, `requestOpenGraph`, and `receiveOpenGraph` actions
- Add OG reducer that stores OG data per URL in `state.openGraph[url]`


mattberridge:
In general I would use `openGraph` not simply `og` as it's clearer? "OG" is an acronym so should be capitalised

sigmarstern:
@mattberridge renamed everything `og` to `openGraph`. Anything else or good to merge?

Tom-Bonnike:
Looks good, can we merge this asap? I need it for the URLs fields.

PR #123, opened 2016-09-28
[RFR] Support Button Logic

mattberridge:
Now we have a proper server-side check for whether the user has already voted (which wasn't working before) - and 3 button states.

screen shot 2016-09-28 at 16 52 45

screen shot 2016-09-28 at 16 52 21

screen shot 2016-09-28 at 16 52 09


mattberridge:
Ping @artibella @sigmarstern @Tom-Bonnike for PR review

src/actions/PetitionActions.js
```diff
@@ -33,6 +33,15 @@ export function fetchPetition (id) {
   };
 }
 
+export function refreshPetition (id) {
```
mattberridge: This is similar to the `fetchPetition` action but does not involve a loading state.
So this happens 'behind the scenes' for the user.

src/containers/Petition.js
```diff
@@ -27,6 +27,10 @@ const PetitionContainer = React.createClass({
       fetchPetition(id).then(({ petition }) => isSupporting
         ? supportPetition(petition)
         : () => {});
+    } else {
+      // Otherwise, refresh the petition
+      // in the background, no loading states
+      refreshPetition(petition.id);
```
mattberridge: Always fetch the petition on the client-side, but the user knows nothing

src/components/PetitionSidebar/index.js
```diff
@@ -13,11 +22,25 @@ const PetitionSidebar = ({ timeMetric, supportable, startDate, runningTime }) =>
...
+          </ButtonIcon>
+        </FakeButton>
+      }
+
+      {!isSupportable &&
+        <FakeButton disabled block>
+          {settings.petitionPage.supportButton.unsupportableText}
+        </FakeButton>
+      }
+    </div>
```
mattberridge: Component is ugly, but I'll refactor it further down the line ;)

src/actions/PetitionActions.js
sigmarstern: Should there also be a test for this function?

PR #124, opened 2016-09-28
[WIP] Sync OG URLs state to form

mattberridge:
@Tom-Bonnike here is what I was thinking, roughly put together anyway.

See any reason why this would not work? Store the links in the state, as they would be used / submitted by the form? basically Sync the redux state and form data


Tom-Bonnike:
Yep ok, makes sense!

src/reducers/openGraph.js
```diff
@@ -12,11 +15,22 @@ export default function openGraph (state = initialState, action) {
...
+      );
     case RECEIVE_OPEN_GRAPH:
+      const appendedLinks = state.links.concat([{
+        url: action.openGraph.url,
+        og: action.openGraph
+      }]);
       return Object.assign({},
         state,
         {
           [action.openGraph.url]: action.openGraph,
```
mattberridge: @Tom-Bonnike is this needed anymore?

PR #137, opened 2016-10-06
Removed unused home route

Tom-Bonnike:
Well, I guess it’s time. Only reason we could still use it is for a direct access to /auth/logout & /auth/login.


mattberridge:
We are adding a home page in this sprint though?

Tom-Bonnike:
Oopsie. :(

PR #139, opened 2016-10-06
Fixed displayed URL of petition in share modal

Tom-Bonnike:


mattberridge:
Think we could add to the helper that the displayed link removes displaying the `http://` / `http://www.` - as we did for the video teasers?

Tom-Bonnike:
I’ll do an helper for that too then. :)

PR #142, opened 2016-10-06
[RFR] Small Petition Changes

mattberridge:


src/components/PetitionResponseStatus/index.js
```diff
@@ -1,22 +1,31 @@
...
-    <div className={styles.top}>
+  <article className={styles.root} >
+    <header className={styles.top}>
       <span className={styles.icon}>
         <Icon id='Flag' size='small' modifier='highlight' />
       </span>
-      <h2>
+      <h2 className={styles.title}>
         {getPetitionResponseStatusTitle(props)}
       </h2>
```
Tom-Bonnike: could be a <h1> in an <article> tag, right?

mattberridge: Not according to what Hugo told me before he left

Tom-Bonnike: Doesn’t change much anyway, was just asking for my personal knowledge but if Hugo says so… :p

PR #144, opened 2016-10-06
[RFR] Basic pageview Tracking implementation

Tom-Bonnike:


src/routes.js
```diff
@@ -1,7 +1,10 @@
+/* eslint no-undef: "off" */
```
Tom-Bonnike: Required as eslint will trigger an error for the undefined property “CustomEvent” in this file. CustomEvent is only defined on the clientside.

src/routes.js
```diff
@@ -12,15 +15,35 @@ import EditPetition from 'containers/EditPetition';
 import PublishedPetition from 'containers/PublishedPetition';
 import RespondToPetition from 'containers/RespondToPetition';
 
+const pageviewEvent = __CLIENT__ &&
+  typeof CustomEvent !== 'undefined' && // Required as it’s not defined during tests.
+  new CustomEvent('IRIS_pageview');
```
Tom-Bonnike: Is it normal that __CLIENT__ is set to true during some tests?

eschaefer:
Ok something is up with Travis since I added `SITE_NAME` to the heroku app's env config, and re-builds don't help.

eschaefer:
Moved to https://github.com/iris-dni/iris-frontend/pull/150

PR #146, opened 2016-10-07
[RFR] Add Trust Route

mattberridge:
Adds `trust/support/:petitionId` route and beginning of Trust flow for **supporting a Petition**.

User can fill out their details:

screen shot 2016-10-10 at 18 29 52

If they log in via the SSO button, the form is pre-filled:

screen shot 2016-10-10 at 18 29 38
 
When proceeding, the user receives a text message to their given phone number. At present, I redirect to the petition page and show the *Thank you for supporting* modal, as the flow would be if the user was *trusted* and needs no further verification. 

**At the moment** this redirect also happens for untrusted users pending building the TAN confirmation flow. Therefore it seems as if the user has voted but the support count will not increase right now. 

## Note
This is a branch for basic functionality. Styling and polishing, validation etc. is still TODO


src/components/PetitionForm/index.js
```diff
@@ -3,7 +3,7 @@ import { reduxForm } from 'redux-form';
 import { createPetition, updatePetition } from 'actions/PetitionActions';
 import petitionValidator from 'form/petitionValidator';
 import Fieldset from 'components/Fieldset';
-import FormField from 'components/FormField';
+import FormFieldsIterator from 'components/FormFieldsIterator';
```
eschaefer: I like this idea. Will probably copy it for the home page modules too.

PR #147, opened 2016-10-07
[RFR] home page modules

eschaefer:
- [x] Add route Home route and container for `/` 
- [x] Add one homepage modules config file, with array of modules to be rendered on HP.
- [x] Add recent petitions container/component (`<HomepagePetitions>`)

Try at https://iris-frontend-dev-pr-147.herokuapp.com/


eschaefer:
ping @mattberridge -- No real styles yet since we don't have them. This just adds a homepage basis.

src/containers/Home.js
```diff
@@ -0,0 +1,56 @@
...
+      const location = Object.assign({}, this.props.location, HOME_QUERY);
+
+      this.props.clearPetitions();
+      this.props.fetchPetitions({ location });
+    }
+  },
+
+  render () {
+    return (
+      <Home {...this.props} />
```
mattberridge: You should include `<Helmet>` for `<head>` management at container level :)

eschaefer: Ah yeah - will add that now ⛑

src/containers/Home.js
```diff
@@ -0,0 +1,55 @@
...
+    );
+  }
+}));
+
+HomeContainer.fetchData = (props) => {
+  const { store } = props;
+  const location = Object.assign({}, props.location, HOME_QUERY);
+
+  return store.dispatch(fetchPetitions({ location }));
```
mattberridge: Still a bit unsure of this but not 100% on the best solution. I think we should make a new action called `fetchLatestPetitions(limit)` or something, which uses `fetchPetitions` inside that. Abstract away the location / logic from the component.

src/containers/Home.js
```diff
@@ -0,0 +1,55 @@
...
+import { withRouter } from 'react-router';
+import { fetchPetitions, clearPetitions } from 'actions/PetitionsActions';
+import Home from 'components/Home';
+import getPetitions from 'selectors/petitions';
+import getPetitionsHomeQuery from 'selectors/petitionsHomeQuery';
+import isClientSideRouting from 'helpers/isClientSideRouting';
+
+const HOME_QUERY = getPetitionsHomeQuery({ limit: 3 });
+
+const HomeContainer = withRouter(React.createClass({
```
mattberridge: I don't think for this, using `withRouter` is needed, that's just when we are grabbing where params on the client-side (for filtering or by `:id`)

mattberridge:
Another comment would be: what happens when we have a "Trending Petitions" panel below too?

Do you think we should assign these petitions to a nested state object i.e.

```
petitions: {
  latest: [],
  trending: []
}
```

?

src/containers/Home.js
```diff
@@ -0,0 +1,55 @@
...
+
+  componentWillMount () {
+    // If there are no petitions, or if the user arrived on the page by clicking
+    // a client-side router link, then we fetch petitions client-side.
+    if (!this.props.petitions.length || isClientSideRouting(this.props.location)) {
```
mattberridge: This block seems a bit overkill here. If we used the separate state objects below we should do a much simpler:

```
if (!this.props.petitions.latest.length) {
  this.props.fetchLatestPetitions();
}
```

src/containers/Home.js
eschaefer: added a new action called `fetchLatestPetitions`

src/containers/Home.js
eschaefer: ah yeah. yanked it out.

src/containers/Home.js
eschaefer: Ya, having the separate state objects like `petitions.latest` and `petitions.trending` would simplify this. I'll take a shot at it

eschaefer: Ok, took a little time, but there's a scalable way to add groupings of petitions now. A helper called `helpers/getPetitionsGroupsList` has an array of groups, like `latest`, `trending`, or whatever else comes. This array determines the added properties to the home page `petitions` state object. Will probably need some tweaking in the future, but this serves what we need rn.

PR #151, opened 2016-10-12
[RFR] Deep merge the configs

eschaefer:
Fixes the issue of some deeply nested properties not getting set if there is a custom config with missing properties in a theme.


mattberridge:
Quickly update the test as well?

`/test/settings/merge.js`

PR #153, opened 2016-10-13
[RFR] Fixing client-side petition groupings

eschaefer:
This just un-breaks the mess I made earlier. Still not great, but I don't want to waste more time on it.


src/containers/Home.js
```diff
@@ -16,8 +16,7 @@ const HomeContainer = React.createClass({
   componentWillMount () {
     // If there are no petitions, or if the user arrived on the page by clicking
     // a client-side router link, then we fetch petitions client-side.
-    if (!hasPetitionGroups(this.props.petitions) || isClientSideRouting(this.props.location)) {
-      this.props.clearPetitions();
+    if (!hasPetitionGroups(this.props) || isClientSideRouting(this.props.location)) {
```
mattberridge: think we can bin the `isClientSideRouting(this.props.location)` from here no?

src/helpers/getPetitionsQueryParams.js
```diff
@@ -4,5 +4,5 @@ export default (params = {}, query = {}) => ({
   cityName: params.cityName || '',
   limit: parseInt(query.limit || 12),
   state: query.state || 'current',
-  sort: query.sort || ''
+  sort: query['sort'] || ''
```
mattberridge: why the change here, out of interest?

src/actions/PetitionsActions.js
```diff
@@ -60,14 +67,18 @@ export function requestPetitions () {
...
 
-export function receivePetitions (petitions, params, qs, group) {
-  const groupedPetitions = group
-    ? { [group]: petitions }
-    : petitions;
+export function receiveGroupedPetitions (petitions, params, group) {
+  return {
+    type: RECEIVE_GROUPED_PETITIONS,
+    params,
+    groupedPetitions: { [group]: petitions }
```
mattberridge: I would still maybe do this in the reducer if possible as it's assigning state

src/selectors/petitionsGroups.js
```diff
@@ -1,12 +1,11 @@
...
 import getPetitions from 'selectors/petitions';
 
-export default (petitions = {}, groups = []) => {
+export default (petitions, groups) => {
   const petitionGroups = {};
 
-  forEach(groups, (group) =>
-    set(petitionGroups, group, getPetitions(petitions[group].data || []))
+  groups.forEach((group) =>
+    set(petitionGroups, group.group, getPetitions(petitions[group.group] && petitions[group.group].data || []))
```
mattberridge: Quite irrelevant but not a fan of the `group.group` naming. But don't change it if there's nothing better :)

mattberridge:
Nicely clean and flexible solution in the end here. A few comments you might examine but overall pretty well good to go

👍

src/actions/PetitionsActions.js
```diff
@@ -29,19 +29,26 @@ export function fetchPetitions ({ location, params }, group) {
...
 }
 
 export function fetchGroupedPetitions (props = {}, groups = []) {
-  const sortQuery = getHomePagePetitionsQuery();
-  const location = Object.assign({}, props.location, sortQuery);
-
   return (dispatch) => {
-    const groupDispatches = groups.map(group =>
-      dispatch(fetchPetitions({ location }, group)));
+    dispatch(requestPetitions());
```
mattberridge: Not sure this logic for `isLoading` is re-useable here actually.
We set it to `petitions: { isLoading: true }` here and then after the first group finishes it's set to `false`?

Do you think we need;

```
latest: {
  isLoading
},
trending: {
  isLoading
}
```

so we could display a loading spinner for each group individually?

Not critical, I guess

src/containers/Home.js
eschaefer: yep

src/helpers/getPetitionsQueryParams.js
eschaefer: My syntax highlighting was showing that `query.sort` was corresponding to the native `sort()` method, so I was OCD and changed it :P

src/actions/PetitionsActions.js
eschaefer: Yep, you're right. Each group needs its own `isLoading` property. And  the `<Loading />` component needs to be wrapped around the homepage petitions...

PR #159, opened 2016-10-14
[RFR] #75 Show connected articles of a petition

artibella:
This shows the article teasers of a petition

can be testet with petition with ID `1lcVp`


src/actions/PetitionActions.js
```diff
@@ -51,7 +51,9 @@ export function requestPetition () {
 }
 
 export function receivePetition (petition) {
+  // TODO: there might be a better place to transform link data
```
eschaefer: Probably do this in the container that maps the petition mentions to props?

PR #175, opened 2016-10-20
[RFR] Fix scroll behavior

eschaefer:
Using https://github.com/taion/react-router-scroll to achieve default browser behavior when changing routes client-side.


eschaefer:
@mattberridge not much to test here, but it works. Route changes scroll to top by default now.

PR #177, opened 2016-10-21
[RFR] Restrict Trust Routes

mattberridge:
Using Higher Order Component conventions we had already used elsewhere (`Restricted` container) and outlined from here:
https://crysislinux.com/limit-access-to-redux-apps-with-higher-order-components/

We restrict access to the `/trust/support/:id` routes if the petition is unsupportable, or if the user has already supported (client-side test)

We restrict access to the `/trust/publish/:id`, `/petitions/:id/edit` and `/petition/:id/preview`routes if the petition is already published.

I quite like this idea of using HOCs to do data fetching and restricting. Means again we can abstract things using the React component lifecycle.

Also note the use of `{ push }` i.e .`dispatch(push)` with `react-router-redux`. We have used elsewhere `withRouter()` and `this.props.router.push`.

### DIFF
--- a/src/containers/RedirectIfUnsupportable.js
+++ b/src/containers/RedirectIfUnsupportable.js
@@ -1,0 +1,51 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import { push } from 'react-router-redux';
+import { fetchPetition, refreshPetition } from 'actions/PetitionActions';
+import getPetition from 'selectors/petition';
+import petitionSupportable from 'selectors/petitionSupportable';
+import petitionUserSupport from 'selectors/petitionUserSupport';
+import petitionPath from 'selectors/petitionPath';
+
+const RedirectIfPublishedWrapper = (WrappedComponent) => {
+  const RedirectIfPublished = React.createClass({
+    componentWillMount () {
+      if (!petitionSupportable(this.props.petition)) {
+        this.props.push(petitionPath(this.props.petition));
+      }
+
+      this.props.refreshPetition(this.props.petition.id)
+        .then(({ petition }) => {
+          if (petitionUserSupport(petition)) {
+            this.props.push(petitionPath(this.props.petition));
+          }
+        });
+    },
+
+    render () {
+      return (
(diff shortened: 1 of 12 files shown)


mattberridge:
Ping @eschaefer @sigmarstern

src/containers/RedirectIfUnsupportable.js
```diff
@@ -0,0 +1,51 @@
...
+import getPetition from 'selectors/petition';
+import petitionSupportable from 'selectors/petitionSupportable';
+import petitionUserSupport from 'selectors/petitionUserSupport';
+import petitionPath from 'selectors/petitionPath';
+
+const RedirectIfPublishedWrapper = (WrappedComponent) => {
+  const RedirectIfPublished = React.createClass({
+    componentWillMount () {
+      if (!petitionSupportable(this.props.petition)) {
+        this.props.push(petitionPath(this.props.petition));
```
eschaefer: Nice to have this in a higher-order component!

src/routes.js
```diff
@@ -41,13 +43,13 @@ export default (props = {}) => (
...
+      <Route path='petitions/:id/preview' component={RedirectIfPublished(PreviewPetition)} />
       <Route path='respond/:token' component={RespondToPetition} />
-      <Route path='trust/support/:id' component={TrustSupport} />
-      <Route path='trust/support/:id/confirm' component={TrustSupportConfirmation} />
-      <Route path='trust/publish/:id' component={TrustPublish} />
-      <Route path='trust/publish/:id/confirm' component={TrustPublishConfirmation} />
+      <Route path='trust/support/:id' component={RedirectIfUnsupportable(TrustSupport)} />
+      <Route path='trust/support/:id/confirm' component={RedirectIfUnsupportable(TrustSupportConfirmation)} />
+      <Route path='trust/publish/:id' component={RedirectIfPublished(TrustPublish)} />
+      <Route path='trust/publish/:id/confirm' component={RedirectIfPublished(TrustPublishConfirmation)} />
```
eschaefer: 👍

PR #178, opened 2016-10-21
[RFR] Owner Relation Display

mattberridge:
**Bug from review**: call petition owner from base object first, then `data` nested object if not present. So non-SSO petitions always display an author name (they don't have nested `data`)

screen shot 2016-10-21 at 12 55 40
screen shot 2016-10-21 at 12 55 24


mattberridge:
Ping @artibella @eschaefer

PR #179, opened 2016-10-21
[RFR] Confirm pre-filled data

mattberridge:
Tweak validation helper to allow valid icon to be shown when forms are pre-filled with data, either after sign-in via SSO or when editing a petition

screen shot 2016-10-21 at 13 18 45


mattberridge:
Ping @eschaefer

PR #180, opened 2016-10-24
[WIP] Petition widget rendering

artibella:
Enables rendering of a petition widget on the `embed/:id` route. Still WIP and can be optimized, but already has it's own JS and CSS bundles.


src/server/render-widget.js
```diff
@@ -0,0 +1,89 @@
+import React from 'react';
```
mattberridge: Think it's diverging too much if we used `preact` for the widget? @artibella @eschaefer?

mattberridge: Also, for me it's a big no to duplicate this file as it's so large / complex. I'll try and think of a way around it.

eschaefer: Would be 100% on board with Preact instead.

artibella: My opinion is that I would not see a huge advantage to switch to Preact for the widget only at this point. Choosing a rendering engine is something that should not differ on a feature/story level but be part of the whole architectural decision.

It would be highly confusing for future contributors to work on a mixed React/Preact code base.

eschaefer:
CSS module names broke somewhere along the way in this setup. Need to figure out what's going on...
screen shot 2016-10-25 at 18 36 26

PR #181, opened 2016-10-24
[RFR] Ad slots

eschaefer:
Pages with wideboard:

- [x] Homepage
- [x] Petition overview/browse
- [x] Petition detail page
- [x] Imprint

Pages with rectangle ad:  

- [x] Petition detail page


src/theme/includes/scripts.html
```diff
@@ -1,9 +1,16 @@
...
 
-       console.log('PAGE VIEW HAPPENED');
-    */
-  });
+    /* Dispatch ad reload when route
+       changes client-side */
+    loadAds();
+  };
+
+  var loadAds = function () {};
```
artibella: Do you think we should have empty implementation examples in the code base of Iris? Maybe consider moving this to the Petitio fork?

src/components/AdContainer/ad-container.scss
```diff
@@ -0,0 +1,29 @@
...
+  }
+}
+
+.rectangle {
+  margin-bottom: 1.25em;
+
+  &:empty {
+    display: none;
+  }
+}
```
artibella: As discussed on Friday I would not include any default styling for ad slot. This should be done by the actual implementation in the theme.

@mattberridge: What is your opinion on this?

src/components/AdContainer/index.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import settings from 'settings';
+import getClassNames from 'helpers/getClassNames';
+import styles from './ad-container.scss';
+
+const AdContainer = ({currentCity = {}, type}) => (
+  <div
+    id={`ad-slot--${type}`}
+    className={getClassNames(styles, ['root', type])}
```
artibella: There should be some static CSS class in place (not CSS modules based) so that external scripts that integrate the advertising functionality can have a simple simple way of selecting all ad slot placeholders: e.g. `document.getElementsByClassName('ad-slot')`. Very similar to what you did one line above with the "id-attribute".

The current CSS classes do not allow this:



So this line could just be something like:

```
className={`ad-slot ad-slot--${type}`}
```

eschaefer: Probably something for diesdas to figure out... I can think of like 3 different ways of doing that same thing w/out classnames.

src/components/AdContainer/ad-container.scss
mattberridge: Yeah, I'd remove `ad-container.scss` entirely

src/components/AdContainer/index.js
mattberridge: As Richard says, just do `className={`ad-slot ad-slot--${type}`}` and be done with it :)

src/components/AdContainer/ad-container.scss
eschaefer: Done

src/components/AdContainer/index.js
eschaefer: Done

src/components/Petition/index.js
```diff
@@ -14,13 +15,17 @@ import ShowWhen from 'components/ShowWhen';
...
-const Petition = ({ preview }) => (
+const Petition = ({ preview, city }) => (
   <article>
     <Section theme={'grey'}>
       <Container>
         <PetitionHeader />
       </Container>
     </Section>
+    <AdContainer
+      currentCity={city.data}
```
eschaefer: https://github.com/iris-dni/iris-frontend/pull/183

src/components/Imprint/index.js
```diff
@@ -6,11 +6,15 @@ import CreateCTA from 'components/CreateCTA';
 import Header from 'components/Header';
 import ImprintDetails from 'components/ImprintDetails';
 import PlatformSupporters from 'components/PlatformSupporters';
+import AdContainer from 'components/AdContainer';
 
 import hasPlatformSupporters from 'helpers/hasPlatformSupporters';
 
 const Imprint = () => (
   <article>
+    <AdContainer
```
eschaefer: Doesn't work for the home page unfortunately. The wideboard doesn't appear at the top of the page there.

PR #183, opened 2016-10-25
[RFR] Use redux container for ad slot

eschaefer:
Moves mapping the `currentcity` data into a container component rather than drilling props into `<AdSlot>` every time it's called.


eschaefer:
@mattberridge @artibella just fyi

PR #188, opened 2016-10-26
[RFR] ImageField component for Redux Form

mattberridge:
Getting `react-dropzone` and `redux-form` to work together, plus adding and removing of the Petition image. Added `RemoveButton` abstraction for the image teasers to click and remove, taken from the links field.

@sigmarstern there is no submitting / uploading of data to the backend yet, this is purely front-end form logic.


mattberridge:
RFR @eschaefer @sigmarstern @artibella

PR #203, opened 2016-11-04
Refactor translation handling

artibella:
Currently all settings and translations are mixed in one file `src/settings/config.js` and this PR is an attempt to separate them in different files so that we can integrate translation keys easier into WebTranslateIt.

Currently rendering breaks for some reasons if `THEME_PATH` ENV variable is defined – need to figure out why some Nunjucks templates can't be included anymore.

@mattberridge: maybe you can have a look at the problem


src/settings/getTranslations.js
```diff
@@ -0,0 +1,13 @@
+import mergeSettings from './merge';
+import translations from 'theme/translations';
+
+const customThemePath = process.env.THEME_PATH || '';
+const useCustomTheme = customThemePath && !process.env.TEST_ENV;
+
+const themeTranslations = useCustomTheme
+  ? require('../../' + process.env.THEME_PATH + '/translations')
+  : {};
```
artibella: should be `require 'theme/translations'`

PR #206, opened 2016-11-07
[RFR] Split settings config

mattberridge:
@artibella adds `setting` and `translation` helpers.

```
import { setting } from 'settings';
import { translation } from 'settings';
```

Huge job to retrofit them into all the places we use `settings.x` though.


mattberridge:
Or do you think it should be split even more?

```
import { translation } from 'translations';
```

PR #243, opened 2016-12-05
[RFM] Responsive Images

mattberridge:


mattberridge:
Ping @sigmarstern think this is finished

PR #259, opened 2017-02-09
[WIP] Email Confirmation: Indicate that an email address has been confirmed…

lukasender:
…, if that's the case


src/client.js
```diff
@@ -20,6 +20,7 @@ import logPageview from 'helpers/logPageview';
 import getHistory from 'helpers/getHistory';
 
 const initialState = window.__INITIAL_STATE__ || {};
+const additionalRenderProps = window.__ADDITIONAL_RENDER_PROPS__ || {};
```
mattboon, line 23: I don't see this as a solution. State should not be in two places.

PR #260, opened 2017-02-09
[RFR] More descriptive error pages for 'Email already confirmed' & 'City Answer already submitted'

mattboon:
# Use Case 1: Unknown response token
screen shot 2017-02-10 at 09 59 04

# Use Case 1: response token already used (City answer submitted)

screen shot 2017-02-10 at 09 58 59

# UseCase 2: Email already confirmed

screen shot 2017-02-10 at 09 59 20


src/containers/EmailConfirmationContainer.js
```diff
@@ -14,4 +15,8 @@ const EmailConfirmationContainer = React.createClass({
   }
 });
 
-export default EmailConfirmationContainer;
+export const mapStateToProps = ({ confirmation }) => ({ confirmation });
```
mattboon: Your `chooseTranslation` logic needs to go here, based on the result of `confirmation`. You might create a `selector` to map the values to a translation string, which would then be testable.

You want to pass down the correct content for the page from the container i.e.
`<EmailConfirmation title={xxx} text={xxx} />`

Always map props in container components and then pass them to "dumb" presentational components e.g. `<EmailConfirmation />` for rendering, rather than adding logic inside those components themselves.

src/server/emailConfirmation.js
```diff
@@ -9,7 +14,27 @@ export default (request, reply) => {
...
       .then((response) => render(request, reply, 'index'))
-      .catch(() => reply('Invalid key').code(500));
+      .catch((error) => {
+        const statusCode = get(error, 'response.data.error.code', 500);
+        switch (statusCode) {
+          case 404:
+            return render(
+              addServerRequestProps(request, { confirmation: { invalidKey: true } }),
+            reply, 'index');
+          case 400:
```
mattboon, line 24:
```
            const description = get(error, 'response.data.error.description', '');
            const confirmationProps = description.indexOf('Already used') > -1
              ? { emailAlreadyConfirmed: true }
              : { displayError: true }

            return render(
              addServerRequestProps(request, { confirmation: confirmationProps }),
            reply, 'index');
```

src/containers/EmailConfirmationContainer.js
```diff
@@ -1,17 +1,29 @@
...
+    const { displayError, translation } = this.props;
+    if (displayError) {
+      return <PetitionResponseTokenError />;
+    }
     return (
       <div>
-        <Helmet title={settings.emailConfirmationPage.title} />
-        <EmailConfirmation />
+        <Helmet title={translation.title} />
+        <EmailConfirmation {...translation} />
```
mattboon: I'd prefer explicit props here as they are clearer

```
<EmailConfirmation
  hint={hint}
  title={title}
  link={link}
/>
```

src/containers/EmailConfirmationContainer.js
```diff
@@ -1,17 +1,29 @@
...
-import settings from 'settings';
+import { connect } from 'react-redux';
 import Helmet from 'react-helmet';
+import getEmailConfirmationResponse from 'selectors/emailConfirmationResponse';
 import EmailConfirmation from 'components/EmailConfirmation';
+import PetitionResponseTokenError from 'components/PetitionResponseTokenError';
 
 const EmailConfirmationContainer = React.createClass({
   render () {
+    const { displayError, translation } = this.props;
```
mattboon: I would lose the nested `translation` object and just pass `title`, `hint` etc. singularly

src/selectors/emailConfirmationResponse.js
```diff
@@ -0,0 +1,11 @@
+import settings from 'settings';
+
+export default ({ displayError, invalidKey, emailAlreadyConfirmed }) => {
+  if (displayError || invalidKey) {
+    return {displayError: true};
+  }
+  if (emailAlreadyConfirmed) {
+    return {translation: settings.emailConfirmationPage.emailAlreadyConfirmed};
+  }
+  return {translation: settings.emailConfirmationPage.default};
```
mattboon: If you lose the `translation` key, you can simplify this block

```
return settings.emailConfirmationPage[emailAlreadyConfirmed ? 'emailAlreadyConfirmed' : 'default']
```

src/server/render.js
```diff
@@ -68,8 +68,10 @@ export default (request, reply, viewName = 'index') => {
       // Set default status code
       let statusCode = 200;
 
+      const serverProps = request.serverProps || {};
```
mattboon: After thinking over lunch, maybe `request.initialState` is clearer :)

mattboon:
Nice @lumannnn - looks great. A few small tweaks but mostly good-to-go

src/server/emailConfirmation.js
```diff
@@ -9,7 +14,27 @@ export default (request, reply) => {
...
+            const description = get(error, 'response.data.error.description', '');
+            if (description.indexOf('Already used') > -1) {
```
mattboon: This might also be a tested helper, like we did here:
https://github.com/iris-dni/iris-frontend/blob/develop/src/helpers/hasSupportedPetitionAlready.js

`hasConfirmedEmailAlready.js` or something

src/server/emailConfirmation.js
lukasender: changed

lusbuab (approved):

PR #262, opened 2017-03-13
Fix stretched logos on imprints page

BennoDev:
- [x] Fix logos being stretched on imprints page
- [x] Fix image requests with decimals in dimensions
- [x] Refactor `mapImageSizeToNewAttrs`

### DIFF
--- a/test/helpers/calculateImageHeight.js
+++ b/test/helpers/calculateImageHeight.js
@@ -14,7 +14,7 @@ describe('calculateImageHeight', () => {
   it('returns an integer', () => {
     const actual = calculateImageHeight(1000, { w: '800', h: '600' });
 
-    assert.typeOf(actual, 'number');
+    assert.equal(actual % 1, 0);
   });
 
   it('fails silently', () => {
--- a/test/helpers/calculateImageHeight.js
+++ b/test/helpers/calculateImageHeight.js
@@ -8,18 +8,17 @@ describe('calculateImageHeight', () => {
     const actual = calculateImageHeight(1000, { w: 800, h: 600 });
     const expected = 750;
 
-    assert.deepEqual(actual, expected);
+    assert.equal(actual, expected);
   });
 
   it('returns an integer', () => {
     const actual = calculateImageHeight(1000, { w: '800', h: '600' });
-    const expected = 750;
 
-    assert.deepEqual(actual, expected);
+    assert.typeOf(actual, 'number');
   });
 
   it('fails silently', () => {
-    const actual = calculateImageHeight();
+    const actual = calculateImageHeight(1000, { h: 1000 });
     const expected = '';
 
     assert.deepEqual(actual, expected);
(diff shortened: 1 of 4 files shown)


BennoDev:
@mattboon @sigmarstern I added the unit tests, does it seem sufficiently covered?

test/helpers/calculateImageHeight.js
```diff
@@ -0,0 +1,27 @@
...
+import calculateImageHeight from 'helpers/calculateImageHeight';
+
+const { assert } = chai;
+
+describe('calculateImageHeight', () => {
+  it('returns the expected height', () => {
+    const actual = calculateImageHeight(1000, { w: 800, h: 600 });
+    const expected = 750;
+
+    assert.deepEqual(actual, expected);
```
sigmarstern: Since you're only comparing integers, I think you could use `assert.equal(actual, expected)`.

test/helpers/calculateImageHeight.js
```diff
@@ -0,0 +1,27 @@
...
+  });
+
+  it('returns an integer', () => {
+    const actual = calculateImageHeight(1000, { w: '800', h: '600' });
+    const expected = 750;
+
+    assert.deepEqual(actual, expected);
```
sigmarstern: Shouldn't we test for `actual` being a number? [`assert.typeOf(actual, 'number')`](http://chaijs.com/api/assert/#method_typeof)

test/helpers/calculateImageHeight.js
```diff
@@ -0,0 +1,27 @@
...
+  });
+
+  it('fails silently', () => {
+    const actual = calculateImageHeight();
+    const expected = '';
+
+    assert.deepEqual(actual, expected);
```
sigmarstern: Is this really failing silently? What about `calculateImageHeight(1000, {h: 1000})`. Might return `NaN`.

test/helpers/calculateImageHeight.js
BennoDev: yeah I was looking for an isInteger function or something in the chai api

sigmarstern: Are there actual integers in JS?

sigmarstern: There's also `assert.isNumber()` in chai.

sigmarstern: To ensure it's an integer you could do `assert.equal(actual % 1, 0)`

test/helpers/calculateImageHeight.js
BennoDev: I've added additional checks to prevent this

test/helpers/calculateImageHeight.js
BennoDev: there are only numbers in javascript, so calculating the rest is probably the best bet

sigmarstern (approved):

PR #263, opened 2017-04-04
Add successful & answered filters

BennoDev:
- [x] Add successful and answered filter
- [x] Add successful but not answered filter

### DIFF
--- a/src/theme/translations.js
+++ b/src/theme/translations.js
@@ -37,6 +37,7 @@ export default {
         "label": "Select a city",
         "placeholder": "Enter name or ZIP"
       },
+      "answered": "Answered petitions",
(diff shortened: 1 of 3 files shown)


src/theme/translations.js
```diff
@@ -37,6 +37,7 @@ export default {
         "label": "Select a city",
         "placeholder": "Enter name or ZIP"
       },
+      "answered": "Answered petitions",
```
BennoDev: We are looking to find a more descriptive term

BennoDev:
The copywriting is not yet properly done, current labels and names are temporary

artibella (approved):

PR #265, opened 2017-04-05
Feature/homepage answered petitions

BennoDev:


src/containers/Home.js
```diff
@@ -41,10 +47,18 @@ const MODULES = [
 
 const PETITION_GROUPS = [
   {
+    group: 'closed',
+    query: {
+      state: 'closed',
+      limit: 3,
+      sort: 'created'
```
artibella: Does this sort by 'created' descending? Maybe there is also a way to sort by answer date (descending) but I am not sure if this available through the API

artibella (commented):
👍  Nice work! Can be merged when copywriting is finalised

PR #266, opened 2017-04-24
Adjust petition filters as discussed

BennoDev:


src/helpers/getPetitionsStates.js
```diff
@@ -1,20 +1,8 @@
 const PETITIONS_STATES = {
-  current: [
+  running: [
     'supportable.active',
     'supportable.winner'
   ],
-  winning: [
```
sigmarstern, line 6: Can we keep those `winning` filter settings and just don't show them as filter options?

BennoDev: Sure, I'll keep the translations aswell

sigmarstern (approved):

PR #274, opened 2017-05-30
Embed Widget: Use 'canonical' as referrer, if present

lukasender:


src/embed.js
```diff
@@ -19,7 +19,16 @@ const fetchHeight = (child) => {
...
+  const linkElements = document.getElementsByTagName('link');
+  const linksArray = Array.prototype.slice.call(linkElements);
+  // get the canonical URL
+  let referrer = null;
+  linksArray.forEach((linkElement) => {
+    if (linkElement.rel === 'canonical' && !referrer) {
+      referrer = linkElement.href;
+    }
+  });
+  const url = getPetitionEmbedUrl(id, referrer);
```
lusbuab: is there a reason to not use `querySelector`: 

```javascript
const canonicalLink = document.querySelector('link[rel="canonical"]')
// canonicalLink is now an element or null
```

lusbuab: http://caniuse.com/#search=querySelector

lukasender: i don't know. don't we have to support older browsers?

lukasender: if not, i'd happily use that one instead

lusbuab: even IE9 supports it: http://caniuse.com/#search=querySelector (click "Show All")

lukasender: very well then :D

lusbuab (approved):

PR #277, opened 2017-10-11
not found /petitions/:id urls redirected to 404-page

natterstefan:
## Story

If user visits `/petitions/:id` with no valid (eg. not found) petition, we redirect him to the 404-page.


src/server/redirects/ProtectRejectedPetition.js
```diff
@@ -10,6 +10,7 @@ export default (nextState, replace, callback) => {
       }
       callback();
     }).catch(() => {
-      callback('Not found');
+      replace('/404');
```
lusbuab, line 14: Why is this in `ProtectRejectedPetition.js`? Is this called on every petition request?

lusbuab: Side Note: A few [lines above](https://github.com/iris-dni/iris-frontend/pull/277/files#diff-4d8f87bf8b04962fb47a003d25eb7550R5) is a potential source of errors:

```javascript
const { params } = nextState || {};
petitionRepository.find(params.id) // <-- params could be undefined
```

lusbuab (approved):

